PR #77, opened 2016-06-14
Adding azure configuration for sending cash on registration

kevinday:
To summarize my understanding from today's discussion in slack

1. fundNewAccount hits our faucet.
2. For simplicity sake an azure install will avoid setting up its own faucet, and instead just sendCash on user registration (which gets created magically).

Does this change look valid?


src/client/accounts.js
```diff
@@ -49,20 +49,25 @@ module.exports = function () {
...
-                    return onFailed(response.statusCode);
-                }
-                onSendEther(account);
-                augur.fundNewAccount({
-                    branch: branch || augur.branches.dev,
-                    onSent: onSent,
-                    onSuccess: onSuccess,
-                    onFailed: onFailed
+            if (process.env.BUILD_AZURE) {
+                var FREEBIE=2.5;
```
kevinday: Is this valid? Or do I need to convert this to FP first?

PR #103, opened 2016-07-11
Add secure login

JohnDanz:
New funding functions in place, updated registration and login flow to utilize the base58 encrypted "secureLoginID"


tinybike:
@JohnDanz I'm looking over this now.  I see that `persist` and `logout` are still part of src/client/accounts.js, and that src/client/db.js still exists.  The sole purpose of db.js is localStorage I/O.  The new accounts model does not use localStorage at all, correct?

tinybike:
Closing this until #106 is added.

Issue #104, opened 2016-07-11
filters.listen hangs when adding an already added filter


kevinday:
Bit of an odd case, but came across this when writing unit tests:

```
augur.filters.listen({
    marketCreated: function (msg) {
        console.log("market created filter:", msg);
    }
}, function (filters) {
	console.log("first complete:", filters);
	augur.filters.listen({
    	marketCreated: function (msg) {
        	console.log("market created filter:", msg);
    	}
	}, function (filters) {
		console.log("second complete:", filters);
	});
});
```
The second filters.listen hangs forever.

tinybike:
Fixed in eb021d82be02a2f6f3d288c57a4d104d5cb6ab72

PR #105, opened 2016-07-11
Allowing multiple listeners per filter type.

kevinday:
@tinybike how do you feel about this change to address #104 that I filed a little earlier?

This does 2 things: 
Allows multiple functions to be registered for a given event. So If I do 

```
augur.filters.listen({
    marketCreated: function1
});
augur.filters.listen({
    marketCreated: function2
});
```
both function 1 and 2 will be registered. Previously if you tried to do this listen just hangs since `next` never gets called and the async.forEachOfSeries completion callback never gets fired.

This also fixes another bug I found where if you do
```
augur.filters.listen({
    foo: function1
});
```
listen hangs forever, again because foo isn't in self.filter and `next` never gets called. With this fix foo just gets ignored.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "augur.js",
-  "version": "1.7.1",
+  "version": "1.7.2",
   "description": "Augur JavaScript API",
   "main": "src/index.js",
   "scripts": {
--- a/src/filters.js
+++ b/src/filters.js
@@ -434,7 +434,7 @@ module.exports = function () {
                 run = async.series;
             }
             async.forEachOfSeries(cb, function (callback, label, next) {
-                if (self.filter[label].id === null && callback) {
+                if (self.filter[label] && callback) {
                     switch (label) {
                     case "contracts":
                         self.start_contracts_listener(function () {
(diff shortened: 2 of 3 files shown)


kevinday:
Ohhh, I like the automated test coverage shaming! :)

(I will add some tests for this before @tinybike merges if he thinks this is a good idea).

kevinday:
Looking over the filter code a bit more it looks like this strategy won't work without some other refactoring. filter[label].id will just get overwritten the second time listen gets called. calling clear_filter on that label won't clear the first id, since it is now effectively orphaned.

Maybe it would be better to first uninstall the filter if it already exists, before reinstalling it with the second call?

tinybike:
Tacking on extra listeners would be consistent with the way Node listeners usually work, although in this case my opinion is that this is a sucky way for listeners to work, especially if (as you point out) you lose the reference to the original filter ID.

I think uninstalling then reinstalling is the cleanest solution.  However, in that case, wouldn't the existing system work ok (just remove then re-add the filter)?  I guess it could be nice to replace a filter without having to manually remove it.  If you'd like to add this feature I'm not opposed to it.

tinybike:
Closing this PR for now.  I'm not convinced this feature is needed and (as per your remark) this PR does not properly implement the feature in any case.  Am open to a PR for replacing existing filters if you think that would be valuable.

kevinday:
The current system does work (uninstalling then reinstalling), though this should happen automatically when you call listen again on the same label (currently it just hangs forever as I mentioned in #104).

I will implement this in another PR.

PR #109, opened 2016-07-12
Filter fix

kevinday:
This addresses the hang I mentioned in #104, and also ignores invalid labels. When a callback is added for a listener that already exists, it is now cleared and replaced by the new one.


tinybike:
Looks good @kevinday!  Merged.

Issue #110, opened 2016-07-14
add tradingFee back to getMarketInfo


kevinday:
Now that cache nodes need to store full market info, it will be helpful if the fields in getMarketInfo are a superset of the fields returned in getMarketsInfo. If this were true, I could construct my full market cache from my calls to getMarketInfo, and get rid of the getMarketInfoCache contract that I wrote (this is how I currently do it except for an additional call to getTradingFee for every market when I initialize the cache, which is slow.

There shouldn't be any performance issues in adding this back, since all the data we need to calculate the tradingFee is already being returned from the chain anyway. Unless there are any objections, I will go ahead and add this back.

tinybike:
Added in f546365f585781700d2f7516462f46e27ff69be8

PR #111, opened 2016-07-15
Trading fee

kevinday:
I'd like to add tradingFee back to getMarketInfo for the reasons mentioned in #110.


tinybike:
Use `tradingFee.toFixed()` instead of re-calling `abi.unfix`.  This function gets called a lot so it's worth optimizing.

kevinday:
Good call. This is now fixed (no pun intended).

Issue #122, opened 2016-07-21
Trading actions: when user doesn't specify limit price treat it as market order


priecint:

tinybike:
Added in 37a951f7f69271c4e97fdb632e7dc71ba0d9212c

PR #123, opened 2016-07-21
Add market order functionality

priecint:
Please focus on whether more test cases are needed

#122


tinybike:
Looks like a good initial set of test cases to me; merging.

tinybike:
Btw small thing the linter picked up is 3 instances of `Use '===' to compare with 'null'` -- already merged this, so will fix on master.

PR #136, opened 2016-07-28
Removed check for max_value, causes undesired behavior when 0

thinkloop:


tinybike:
Good catch!  Sloppy on my part.  Merging.

PR #146, opened 2016-09-30
594 fix cache node fallback

JohnDanz:
https://app.clubhouse.io/augur/story/594/fix-cache-node-fallback

This change simply empties the `augurNodes.nodes` array if there is an error connecting to a cache node. This will stop the code from retrying the cache node since it doesn't have any to check for anymore and fall back to the hosted node connection to load markets in chunks as expected.

### DIFF
--- a/src/modules/compositeGetters.js
+++ b/src/modules/compositeGetters.js
@@ -71,11 +71,9 @@ module.exports = {
             return this.loadMarketsHelper(branchID, chunkSize, isDesc, chunkCB);
         }
         this.augurNode.getMarketsInfo(branchID, function (err, result) {
-						console.log(err);
-						if (typeof result === 'string')
-							console.log(result.charAt(0));
             if (err) {
                 console.warn("cache node getMarketsInfo failed:", err);
+								// Abandon the use of cache nodes since there was a connection issue.
 								self.augurNode.nodes = [];
                 // fallback to loading in batches from chain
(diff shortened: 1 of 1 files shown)


kevinday:
Why would we clear all nodes if only one fails? Augur.js should be able to handle nodes coming and going with out knowing everything away after a single failure.

tinybike:
At the moment there is only 1 node, and it fails quite un-gracefully.  This is ok as a quick fix to keep the site from randomly faceplanting.  @JohnDanz is there a reason we can't just pop the bad node off the node array?

Prob also want load-balancing between nodes.  If you go back 6 months or so, ethrpc had logic for failing node removal, with separate logic for local and remote nodes + load balancing + best-node-selection logic -- could prob copy-paste a lot of that for augur-nodes.

tinybike:
Btw @kevinday if you have the time/inclination to make a PR for load balancing / node selection / etc. it would be much appreciated!

kevinday:
OK cool, yeah I have a few other outstanding augur_node issues I will file
too before I forget about them. Also, I have the documentation updated but
let me know if anything is unclear. Augur.js is still using the old dump
all market data at once cache node, so it will have to be updated to work
with the new elasticsearch nodes.

(I'd be happy to do a call at some point if you think it would be helpful
to walk through all this, give a docker primer, etc)

PR #154, opened 2017-01-12
1844 indentation

stephensprinkle:


stephensprinkle:
Let me fix the `unverified` sig for the commits.  I’ll re-open the commit in just a moment.

PR #164, opened 2017-03-07
fixed overlooked hardcoded to fields

JohnDanz:
Overlooked to fields in a transaction that had been hardcoded in a few assertions for trade.trade and trade.short_sell.


JohnDanz:
I don't know why this is saying coverage goes down, on my local machine i'm getting 96.1%. I will see if i can see why...

PR #165, opened 2017-03-07
Raw transactions

tinybike:


tinybike:
Note: depends on https://github.com/AugurProject/ethrpc/pull/14

PR #168, opened 2017-04-18
Bugfixes found while trying to get Augur working with Parity PoA chain

MicahZoltu:
Fixes duplicate ethrpc inclusion:  ethrpc needs to come from `ethereumjs-connect`, otherwise you end up using/mutating a copy that is never connected and should never be used.

Removes timeout from getLogs object:  This is not used by ethrpc, and it is not cleared by ethrpc because it is on a nested object rather than the params array directly, and it causes problems in Parity since Parity doesn't like extraneous parameters.  Same fix as a previous issue with `timeout` being added to a sub-object.

Fixes a bug where the last chunk of markets wouldn't load:  `startIndex` for the last chunk worth of markets will always be 0.  If sort is descending, this will result in `numMarketsToLoad` being set to `0` which terminates the load prematurely.

### DIFF
--- a/src/chat.js
+++ b/src/chat.js
@@ -4,7 +4,7 @@ var abi = require("augur-abi");
 
 module.exports = function () {
 
-  var rpc = this.rpc;
+  var augur = this;
 
   return {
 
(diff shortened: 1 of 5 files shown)


tinybike:
16 failing tests in test/unit/core/logs.js

tinybike:
@MicahZoltu looks like there's still one broken test, went ahead and merged though, can you fix?

PR #171, opened 2017-06-01
switching getMarketPriceHistory to use chunked logs

kevinday:
Not quite ready to be merged, but PR-ing for review.

A few tests are failing, but seems to be a problem with get-chunked-logs and not these changes specifically.
`rpcInterface.getCurrentBlock()` is returning null here when the tests get unhappy about when trying to call `number` on:
https://github.com/AugurProject/augur.js/blob/master/src/logs/get-logs-chunked.js#L20

Do you know why this cull would be returning null?


tinybike:
Hm, no I don't -- is `rpcInterface.getCurrentBlock()` returning `null` for you in actual use, or just in the tests?  (If just in the tests, did you include that in your mock?)

kevinday:
AFK for a bit but I will double check later.

kevinday:
The issue was that I wasn't mocking that call. John showed me how to do that and the getMarketPriceHistory are now passing again. Hmm, Travis is showing a createBranch test failing which I didn't touch, and all tests pass locally for me 🤔

kevinday:
@tinybike looks like tests are passing now. Can you review this when you get a chance?

tinybike:
Looks good, thanks!

PR #172, opened 2017-09-26
Features/3673 max fee

pgebheim:
Expose maxSettlementFee from simulate trade API to return the worst case fees associated with a proposed entry on the order book.


src/trading/simulation/simulate-buy.js
```diff
@@ -32,6 +34,10 @@ function simulateBuy(outcome, sharesToCover, shareBalances, tokenBalance, userAd
     }
   }
 
+
+  // Worst-case settlement fee for the buyer is if they buy at minPrice
+  simulatedBuy.maxSettlementFees = calculateSettlementFee(sharesToCover, marketCreatorFeeRate, maxPrice.minus(minPrice), shouldCollectReportingFees, reportingFeeRate, minPrice);
```
tinybike: The price at which the buyer is buying should be set by the prices of the orders they're buying, not `minPrice`.  `minPrice` is the minimum possible price allowed on this market.

src/trading/simulation/simulate-sell.js
```diff
@@ -32,6 +34,11 @@ function simulateSell(outcome, sharesToCover, shareBalances, tokenBalance, userA
     }
   }
 
+  // Worst-case settlement fee for the seller is the max market price is hit and
+  // the seller must pay all the fees. This is equivilent to the taker short price going
+  // to 0, which we can calculate using calculateSettlementFee
+  simulatedSell.maxSettlementFees = calculateSettlementFee(sharesToCover, marketCreatorFeeRate, maxPrice.minus(minPrice), shouldCollectReportingFees, reportingFeeRate, 0);
```
tinybike: The goal is not to calculate the maximum possible settlement fee for this market, but rather, *if* the user is creating (making) a new order, to calculate what the maximum possible settlement fee they *could* end up paying *for that particular order*.

tinybike:
Closing this for now as the correct quantity is not being calculated; see comments.

PR #173, opened 2017-09-27
Fix wrong # of shares being used in simulation

pgebheim:
We were originally using the full amount of shares for the ask order, instead of only the amount over which was available in the take simulation.

Do these new token depletion values look right?


tinybike:
Not quite, pushed a fix and closing this PR (it's on a fork so my push didn't modify the PR).

pgebheim:
Got it, took a look at your changes.

PR #175, opened 2017-10-06
event signature needs arguments

epheph:
@tinybike


epheph:
I wonder if we should actually pass in arguments as second argument array so we're not doing string concatenation in calling functions?

PR #176, opened 2017-10-24
Feature/4237 subscriptions

pgebheim:
Add augur-node specific event subscriptions.

Do we want this more integrated into the normal log subscriptions?

### DIFF
--- a/src/augur-node/dispatch-json-rpc-response.js
+++ b/src/augur-node/dispatch-json-rpc-response.js
@@ -6,7 +6,7 @@ var isObject = require("../utils/is-object");
 
 function dispatchJsonRpcResponse(err, jsonRpcResponse) {
   if (err) throw err;
-  if (!jsonRpcResponse || !isObject(jsonRpcResponse) || jsonRpcResponse.id === undefined ) {
+  if (!jsonRpcResponse || !isObject(jsonRpcResponse) || jsonRpcResponse.id === undefined) {
     throw new Error("Bad JSON RPC response received:" + JSON.stringify(jsonRpcResponse));
   }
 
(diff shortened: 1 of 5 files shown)


src/augur-node/dispatch-json-rpc-response.js
```diff
@@ -6,17 +6,28 @@ var isObject = require("../utils/is-object");
...
+  var callback, result;
+  if (jsonRpcResponse.id !== null) {
+    callback = augurNodeState.getCallback(jsonRpcResponse.id);
+    augurNodeState.removeCallback(jsonRpcResponse.id);
+    result = jsonRpcResponse.result;
+  } else if (jsonRpcResponse.result.subscription !== null) {
+    callback = augurNodeState.getCallback("event:" + jsonRpcResponse.result.subscription);
+    result = jsonRpcResponse.result.result;
+    console.log(callback, result);
+  }
```
tinybike: Should there be an `else` here with "Bad JSON response received"?

src/augur-node/subscribe-to-event.js
```diff
@@ -0,0 +1,18 @@
...
+var augurNodeState = require("./state");
+var submitJsonRpcRequest = require("./submit-json-rpc-request");
+
+function subscribeToEvent(eventName, params, callback) {
+  params = Array.isArray(params) ? params : [];
+  params.unshift(eventName);
+
+  submitJsonRpcRequest("subscribe", params, function (err, response) {
+    if (err) throw new Error("Error subscribing to event: " + err);
+    augurNodeState.setCallback("event:" + response.subscription, callback);
```
tinybike: It might be cleaner to have separate `set` and `get` functions for subscriptions, vs always doing string concatenation with `"event:"`.

src/augur-node/dispatch-json-rpc-response.js
pgebheim: This should be handled above with that big IF block, but I think its cleaner to do in here. Will change

PR #180, opened 2017-11-15
Add worstCaseFees to simulatedTrade output

pgebheim:


tinybike:
`simulateTrade` has some tests that need to be updated to include the new field.

tinybike:
Lovely

PR #181, opened 2017-11-16
Error Message Correction

JohnDanz:
While working on initial liquidity in create-market of the UI, I noticed the error message for the `augur.trading.simulateTrade` function was incorrect. It's checking whether `orderType` is equal to `0` or `1` but then if it's not, it throws an error that claims the value should be `1` or `2`. I have corrected the text to properly reflect the values `orderType` is actually looking for.


tinybike:
Good catch.  You need to also update the tests though, travis currently failing.

PR #186, opened 2017-11-22
Change parse-block to emit actual block object, not just number

epheph:
/cc @MicahZoltu for type sanity

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,8 @@
 /// <reference types="node" />
 
+import { Block } from 'ethereumjs-blockstream'
+export { Block }
+
 type AbiEncodedData = string;
 type Address = string;
 type Bytes32 = string;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -1,5 +1,8 @@
 /// <reference types="node" />
```
MicahZoltu: Unrelated to this PR, but this comment is the old way of doing this and shouldn't be included anymore I believe.

epheph: Want me to yank while i'm in there?

MicahZoltu: Totally your call.  I could be wrong, so you should set aside some time to make sure yanking it doesn't break anything if you do.  :)

epheph:
Coverage is flagging this PR because i reduced code in a covered function?

MicahZoltu:
😆 🤣

PR #191, opened 2017-12-05
4911 human readable shares

tinybike:


tinybike:
Implements suggested scheme in CH 4911; @joeykrug if you have time please glance over this and see if I'm messing anything up...

PR #192, opened 2017-12-10
Adding/updating JSDoc documentation

adrake33:
Adding/updating documentation for Augur Node functions so I can use JSDoc to generate documentation for functions in the Simplified API.


src/accounts/get-account-transfer-history.js
```diff
@@ -1,7 +1,37 @@
...
 
+/**
+ * @typedef {Object} AccountTransfer
+ * @property {string} transactionHash Hash returned by token transfer.
+ * @property {number} logIndex Number of the log index position in the Ethereum block containing the transfer.
+ * @property {number} creationBlockNumber Number of the Ethereum block containing the transfer.
+ * @property {string} blockHash Hash of the Ethereum block containing the transfer.
+ * @property {number} creationTime Timestamp, in seconds, when the Ethereum block containing the transfer was created.
+ * @property {string} sender Ethereum address of the token sender.
+ * @property {string} recipient Ethereum address of the token recipient.
```
epheph: Worth mentioning that it `sender` or `recipient` could be null, indicating a burn or a mint?

src/markets/get-disputable-markets.js
```diff
@@ -1,7 +1,22 @@
+/**
+ * @todo Update function description & return information once function has been implemented.
+ */
+
 "use strict";
 
 var augurNode = require("../augur-node");
 
+/**
+ * This function has not been implemented yet. Returns the markets in a specific reporting window that are able to be disputed, along with the value of the dispute bond needed to dispute each market’s result. Requires an Augur Node connection.
```
epheph: Given "Dispute-Only Reporting System" this is going to change pretty significantly

src/accounts/get-account-transfer-history.js
adrake33: Ah, good point.  Yeah, I'll add that.

src/markets/get-disputable-markets.js
adrake33: Yeah, that's fine. This can be left as-is for now and updated later.

src/markets/get-markets-awaiting-reporting.js
```diff
@@ -3,12 +3,17 @@
...
  * @param {Object} p Parameters object.
- * @param {string=} p.universe Look up markets within this universe.
- * @param {string=} p.reportingWindow Look up markets within this reportingWindow
- * @param {string=} p.reportingRound Look up markets in this reportingState: FIRST_REPORTING|LAST_REPORTING. See REPORTING_STATE enum
- * @param {function} callback Called when reporting history has been received and parsed.
- * @return {Object} Markets history, keyed by market ID.
+ * @param {string=} p.universe Contract address of the universe from which to retrieve markets, as a hexadecimal string. If this parameter is not specified, a reporting window must be specified instead.
+ * @param {string=} p.reportingWindow Contract address of the reporting window from which to retrieve the markets, as a hexadecimal string. If this parameter is not specified, a universe must be specified instead.
+ * @param {string=} p.reportingState Reporting state name. Valid values are "DESIGNATED_REPORTING", "FIRST_REPORTING", or "LAST_REPORTING".
```
epheph: This will definitely change, as there's no such thing as first/last

src/reporting/get-reporting-history.js
```diff
@@ -1,14 +1,62 @@
...
+ * @typedef {Object} Report
+ * @property {string} transactionHash Hash to look up the reporting transaction receipt.
+ * @property {number} logIndex Number of the log index position in the Ethereum block containing the reporting transaction.
+ * @property {number} creationBlockNumber Number of the Ethereum block containing the reporting transaction.
+ * @property {string} blockHash Hash of the Ethereum block containing the reporting transaction.
+ * @property {number} creationTime Timestamp, in seconds, when the Ethereum block containing the reporting transaction was created.
+ * @property {string} marketID Contract address of the market, as a hexadecimal string.
+ * @property {REPORTING_STATE} marketReportingState Reporting state of the market.
+ * @property {string} reportingWindow Reporting window the market is in currently.
+ * @property {Array<number|null>} payoutNumerators Array representing the payout set.
```
epheph: we can drop the `|null` . In the response object, we'll have pruned out the null's when the number of outcomes is <8

src/reporting/get-reporting-history.js
```diff
@@ -1,14 +1,62 @@
...
+ * @property {number} amountStaked Description the reporter has staked on the outcome of their report.
+ * @property {string} stakeToken Contract address of the stake token, as a hexadecimal string.
+ * @property {boolean} isCategorical Whether the market is categorical.
```
epheph: Reading these, i wonder if the names are confusing. `isCategorical` and `isScalar` are fixed attributes of the market, while `isIndeterminate` is related to outcome. @tinybike thoughts about changing these? Perhaps making it `isIndeterminate` `payoutIndeterminate`?

src/reporting/get-reporting-history.js
```diff
@@ -1,14 +1,62 @@
...
+ * @param {number=} p.earliestCreationTime Earliest timestamp, in seconds, at which to truncate history results. (This timestamp is when the block on the Ethereum blockchain containing the report submission was created.)
+ * @param {number=} p.latestCreationTime Latest timestamp, in seconds, at which to truncate history results. (This timestamp is when the block on the Ethereum blockchain containing the report submission was created.)
+ * @param {string=} p.sortBy Field name by which to sort the reporting history.
+ * @param {boolean=} p.isSortDescending Whether to sort the reporting history in descending order by sortBy field.
+ * @param {string=} p.limit Maximum number of reporting history reports to return.
+ * @param {string=} p.offset Number of reporting history reports to truncate from the beginning of the results.
  * @param {function} callback Called when reporting history has been received and parsed.
- * @return {Object} Reporting history, keyed by market ID.
+ * @return {Object} Reporting history, keyed by universe, market ID, or reporting window. Each report is of type {@link Report}.
```
epheph: It's just `universe, marketID`

src/markets/get-markets-awaiting-reporting.js
adrake33: I suppose I can change it to "Description pending." for now.

src/reporting/get-reporting-history.js
adrake33: Sounds good.  I'll remove the `|null`.

src/reporting/get-reporting-history.js
adrake33: Good catch! Will update this.

adrake33:
Thanks for reviewing this, @epheph!  Do you have any info on how to calculate any of these:

category popularity
outstanding shares
outcome volume
market volume
UserTradePosition.numSharesAdjustedForUserIntention (what is this?)
UserTradePosition.realizedProfitLoss
UserTradePosition.unrealizedProfitLoss

epheph:
outstanding shares - Since they are all sold in complete sets, querying any staketoken gives you a the outstanding shares (as all stake tokens have the same supply)

category popularity, market volume, and outcome volume are all closely related.

outcome volume goes up and down, based on the trades that are happening on that outcome.
outcome volumes sum's up to the market's volume
market volume sum's up to the category's volume

This file is where the magic happens for those attributes:
https://github.com/AugurProject/augur-node/blob/master/src/blockchain/log-processors/order-filled/update-volumetrics.ts

epheph:
OK, as for `UserTradePosition.numSharesAdjustedForUserIntention`. I haven't unwound this code before, but i took a stab at it. This file is the relevant bit (that's deeply nested)

https://github.com/AugurProject/augur.js/blob/138ed6636ed3e9cd5da01de3e297fd08b711d72a/src/trading/profit-loss/longer-position-pl.js#L12

It seems like it's an accounts total stake on an outcome, when you add the longs and substract the shorts. @tinybike could you add more context about the "Adjusted for intention"?

adrake33:
@epheph Ok, so in other words, sharesOutstanding is the number of complete sets, not the number of shares * number of outcomes, correct?

From what's in `update-volumetrics.ts`, it looks like:
amount = numCreatorShares + numCreatorTokens / price
market volume = previouslyCalculatedMarketVolume + amount
outcome volume = previouslyCalculatedOutcomeVolume + amount
category popularity = previouslyCalculatedPopularity + amount

However, you said the outcome volume can go up or down. So that means `amount` can sometimes be negative? What would cause that to be the case? Aren't `numCreatorShares`, `numCreatorTokens`, & `price` always positive? Perhaps I'm not understanding what they represent.

PR #193, opened 2017-12-15
Augur event emitter with disconnect/reconnect events

pgebheim:


src/augur-node/index.js
```diff
@@ -1,9 +1,10 @@
 "use strict";
+var ee = require("event-emitter");
```
tinybike: Suggest using `var EventEmitter` instead of `var ee`.  Even though this file is trivial it's best not to get in the habit of using overly-brief variable names.

src/connect.js
```diff
@@ -6,7 +6,7 @@ var ethrpc = require("ethrpc");
 var contracts = require("./contracts");
 var api = require("./api");
 var rpcInterface = require("./rpc-interface");
-var connectToAugurNode = require("./augur-node").connect;
+var augurNode = require("./augur-node").connect;
```
tinybike: I think this needs to be just `var augurNode = require("./augur-node");` since you reference `augurNode.connect` below.

src/connect.js
```diff
@@ -76,6 +82,12 @@ function connect(connectOptions, callback) {
         ethereumConnectionInfo.contracts = ethereumConnectionInfo.contracts || contracts.addresses[DEFAULT_NETWORK_ID];
         self.api = api.generateContractApi(ethereumConnectionInfo.abi.functions);
         self.rpc = rpcInterface.createRpcInterface(ethereumConnectionInfo.rpc);
+        ethereumConnectionInfo.rpc.getTransport().addReconnectListener(function () {
+          rpcInterface.emit("reconnect");
```
tinybike: I don't think `rpcInterface` has an `emit` method -- should this be `augurNode.emit` instead?  (Same comment for the disconnect listener callback.)

env.js
```diff
@@ -42,3 +42,16 @@ augur.connect({ ethereumNode, augurNode }, (err, connectionInfo) => {
     });
   }
 });
+
+augur.augurNode.on('disconnect', function() {
```
tinybike: Strings should use double-quotes.  Does the linter not complain about this? :thinking:  I wonder if we're accidentally skipping a directory or something.

src/rpc-interface.js
```diff
@@ -1,4 +1,5 @@
 "use strict";
+var ee = require("event-emitter");
```
tinybike: Same comment as above, prefer to avoid obscure variable names like `ee` if possible.

src/connect.js
pgebheim: rpcInterface is the ethrpc interface, I mixin EventEmitter there as well.

env.js
pgebheim: This dir isn't linted. Fixed.

PR #195, opened 2017-12-18
instead of the undefined error when talking to an unknown network

epheph:


epheph:
@pgebheim  working on fixing tests, merging this

PR #196, opened 2017-12-18
refactor immediateFill/placeTrade

epheph:


src/trading/place-trade.js
```diff
@@ -28,34 +28,24 @@ var constants = require("../constants");
...
+
 function placeTrade(p) {
   var normalizedPrice = normalizePrice({ minPrice: p.minPrice, maxPrice: p.maxPrice, price: p.limitPrice });
   var numTicks = new BigNumber(p.maxPrice, 10).minus(new BigNumber(p.minPrice, 10)).dividedBy(new BigNumber(p.tickSize, 10)).toFixed();
+  var orderType = (["buy", "sell"])[p._direction];
   getBetterWorseOrders({
     marketID: p._market,
     outcome: p._outcome,
-    amount: p.amount,
+    orderType,
```
epheph: Should we make getBetterWorse accept the 0/1 ? we already had some heartache with bid/ask buy/sell

epheph:
tests pass, there are 2, but i don't think they are testing different things now, looking into it

src/trading/place-trade.js
tinybike: Hm, I'd say keep it as "buy"/"sell" for augur-node, since that's what's used everywhere in augur-node -- best to be consistent IMO

src/trading/place-trade.js
```diff
@@ -28,34 +28,24 @@ var constants = require("../constants");
...
     price: normalizedPrice,
```
tinybike: Unless `getBetterWorseOrders` denormalizes the price in augur-node, this should be `price: p.limitPrice` instead of `price: normalizedPrice`.

PR #198, opened 2017-12-19
Mocked data improvements

tinybike:


tinybike:
Note: this PR works with https://github.com/AugurProject/augur-contracts/commit/798472ad6daeabc919cc35d41dfed1b29895c514 -- holding off on updating the augur-contracts version in package.json for now b/c it seems like the versioning is a bit of a mess!

epheph:
Great approach, should save a lot of hearth-ache for future changes and something we should push more widely. Any reason not to tackle the indexed fields too?

tinybike:
Not really -- just did the non-indexed ones because converting between the serialized log data and the human-readable parameters is sucky.  The indexed ones are easier to read, they're just int256s.

epheph:
Sure, there's also fewer of them and they are less likely to change.

tinybike:
Update: augur.js tests are now congruent with https://github.com/AugurProject/augur-contracts/blob/augur-core/feature/deploy_aura

src/format/log/format-log-message.js
```diff
@@ -16,11 +17,17 @@ function formatLogMessage(contractName, eventName, message) {
...
             extraInfo = null;
           }
-          return assign({}, message, {
+          var formattedMessage = {
             extraInfo: extraInfo,
             marketCreationFee: speedomatic.unfix(message.marketCreationFee, "string"),
             topic: decodeTag(message.topic),
-          });
+          };
+          if (message.marketType === 1) {
```
epheph: surprised we don't have a constant for this, perhaps i'll add it in a separate PR.

tinybike: Agree, that would be good!

tinybike:
Going ahead and merging this even though travis isn't passing -- failure is just b/c the tests need to be run vs the deploy_aura augur-contracts (verified this locally).

PR #214, opened 2018-01-16
getUnclaimedMarketCreatorFees

epheph:


epheph:
@adrake33 i finally wrote this thing you asked for. Could i get some help on the jsdoc format for an object return with dynamic keys (marketIDs?)

adrake33:
Does it make sense to return it as an object?  It seems like technically it should be an array if it's keyed by marketIDs.  If it should be an object, I think you could do something like: 

```@returns {Object.<string, number>}```

**Comment by epheph** [comment]
@adrake33 perhaps it does, it would be more like how getMarketsInfo works.

@stephensprinkle do you have a preference?

```
{
        "0x0000000000000000000000000000000000000001": 0,
        "0x0000000000000000000000000000000000000002": 0,
      }
```
vs
```
[{
        marketID: "0x0000000000000000000000000000000000000001",
        unclaimedFee: 0
},{
        marketID: "0x0000000000000000000000000000000000000002",
        unclaimedFee: 0
      }
]
```

stephensprinkle:
Pref would be 1st for me.

epheph:
@stephensprinkle alright, that's how it's already written!

epheph:
Updated jsdoc

adrake33:
@epheph It looks like Sprinkle edited his original comment. Can you please update it to be an array?

epheph:
https://github.com/AugurProject/augur-node/pull/160/files

PR #215, opened 2018-01-18
return event subscription in onComplete callback for later unsubscription

epheph:


epheph:
Checks failed because coveralls is down

PR #217, opened 2018-01-19
Simple helper scripts

bthaile:
created helper scripts to:
1. create open orders
2. fill specific buy/sell orders on market outcome
3. do initial report (for DR or other user)
4. check market creator mailbox
5. list markets and give indication if endDate has passed
6. finalize market
7. set the time by passing in unix timestamp


scripts/helpers/do-initial-report.js
```diff
@@ -0,0 +1,50 @@
...
+        meta: auth,
+        tx: { to: marketID  },
+        _payoutNumerators: payoutNumerators,
+        _invalid: invalid,
+        onSent: function (result) {
+          console.log(chalk.yellow.dim("Sent:"), chalk.yellow(JSON.stringify(result)));
+        },
+        onSuccess: function (result) {
+          console.log(chalk.green.dim("Success:"), chalk.green(JSON.stringify(result)));
+          process.exit(1);
```
epheph: tiny deal, but exit codes are backwards, usually want to exit 0 on success and exit >0 on error

scripts/rep-faucet.js
```diff
@@ -15,7 +15,7 @@ function faucetInAndMigrate(augur, universe, auth, callback) {
     augur.api.ReputationToken.balanceOf({ _owner: auth.address, tx: { to: reputationToken } }, function (err, reputationTokenBalance) {
       if (err) return callback(err);
       console.log("reputationTokenBalance:", reputationTokenBalance);
-      if (new BigNumber(reputationTokenBalance, 10).gt(new BigNumber("100000", 10))) return callback(null);
```
epheph: Might be more correct to make this new `Error("Already loaded")` but not too big a deal for a one-off script

bthaile: I'll make the change

bthaile:
added a few more scripts, the big one is to create lots of liquidity on a market.

PR #220, opened 2018-01-22
Feature/include artifacts

pgebheim:
Include build artifacts in augur.js instead of augur-contracts repo


epheph:
One test references augurContracts: `test/unit/create-market/mock-market-created-data.js`

PR #221, opened 2018-01-22
getFinalizationTime is on Market. also added doc for need _shareHolde…

bthaile:
doc changes and getFinalizationTime is on Market class.


epheph:
Seems important 👍  good catch

PR #228, opened 2018-02-06
new states for new dispute system

epheph:
@adrake33 @nuevoalex


src/constants.js
```diff
@@ -11,16 +11,13 @@ module.exports = {
...
-    LAST_REPORTING: "LAST_REPORTING",
-    LAST_DISPUTE: "LAST_DISPUTE",
-    FORKING: "FORKING",
-    AWAITING_FINALIZATION: "AWAITING_FINALIZATION",
+    OPEN_REPORTING: "OPEN_REPORTING",
+    CROWDSOURCING_DISPUTE: "CROWDSOURCING_DISPUTE",
+    AWAITING_NEXT_WINDOW: "AWAITING_NEXT_WINDOW",
     FINALIZED: "FINALIZED",
+    FORKING: "FORKING",
+    AWAITING_NO_REPORT_MIGRATION: "AWAITING_NO_REPORT_MIGRATION",
```
adrake33: What would this status be used for in Augur Node? After chatting with @nuevoalex on Discord, this seems like it might not be needed, since initial reports can be filed at any time after a fork occurs.

epheph: The reason is because we use this state to store what are actually several different variables. It's making me wonder if forking state should be a separate field. 

If we think it's important to remember that something is "awaiting_fork_migration" we need a way to indicate that. But we also need a way to indicate that this market has had no report from its designated reporter, and now is ready for open reporting.

Thoughts on splitting forking state into something else? Things will basically always follow the same state-pattern , and there is a separate forkingState with the possibilities: "forking", "awaiting migration"

adrake33: I suppose a separate FORKING_STATE would make sense.  REPORTING_STATE makes me think of the states we have listed in this diagram: http://docs.augur.net/#market-state-progression.

epheph: forking is confusing. because it's sometimes something a market DOES (as outlined in that diagram) and sometimes something that happens to a market, which could be in a variety of other states. i'll think on this some more

PR #235, opened 2018-02-10
Flash helper scripts

bthaile:
initial move from individual scripts to move time when USE_NORMAL_TIME = 'false' when uploading contracts either in local env or on test net.

compiling helper scripts into one place following the dp scripting model.

added help for overall running on networks (from dp) and help by each command script.


scripts/flash/designated-report.js
```diff
@@ -0,0 +1,87 @@
...
+
+"use strict";
+
+var chalk = require("chalk");
+var getTime = require("./get-timestamp");
+
+/**
+ * Move time to Market end time and do initial report
+ */
+function designateReportInternal(augur, marketId, outcome, invalid, auth) {
```
tinybike: Nitpick: suggest using `marketID` instead of `marketId` for consistency, and pick one of `outcome` or `outcomeID` (not `outcomeId`) but do not use both `outcome` and `outcomeId` to refer to the same thing.

bthaile: Will do, we are moving to `Id` in UI, I'll change to `ID` for augur.js

tinybike: If you're using `Id` in the UI, that's fine to use it here, too.

scripts/flash/initial-report.js
```diff
@@ -0,0 +1,70 @@
...
+
+function initialReport(augur, params, auth, callback) {
+  if (!params || params === "help" || params.split(",").length < 3) {
+    help(callback);
+  } else {
+    var paramArray = params.split(",");
+    var invalid = paramArray.length === 4 ? paramArray[3] : false;
+    var marketID = paramArray[0];
+    var outcomeId = paramArray[1];
+    var userAuth = getPrivateKeyFromString(paramArray[2]);
```
tinybike: Recommend using the same conventions employed elsewhere in the deployment scripts: `ETHEREUM_PRIVATE_KEY` environment variable for the private key instead of a positional argument.

scripts/flash/set-timestamp.js
```diff
@@ -0,0 +1,30 @@
...
+  augur.api.TimeControlled.setTimestamp({
+    meta: auth,
+    tx: { to: address },
+    _timestamp: numberTicks,
+    onSent: function (result) {
+      console.log(chalk.yellow.dim("Sent Change Time"), chalk.yellow(JSON.stringify(result)));
+      console.log(chalk.yellow.dim("Waiting for reply ...."));
+    },
+    onSuccess: function (result) {
+      var timestamp = augur.api.Controller.getTimestamp();
```
tinybike: Suggest providing a callback to `augur.api.Controller.getTimestamp`, so that this request is not blocking (synchronous).

scripts/flash/get-timestamp.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var encodeTag = require("../../src/format/tag/encode-tag");
+
+function getTime(augur, auth, callback) {
+  var timestamp = augur.api.Controller.getTimestamp();
```
tinybike: Needs callback for async.

scripts/flash/get-timestamp.js
```diff
@@ -0,0 +1,19 @@
...
+  var controller = augur.contracts.addresses[augur.rpc.getNetworkID()].Controller;
+
+  augur.api.Controller.lookup({ meta: auth, tx: {to: controller}, _key: encodeTag("Time")}, function (err, timeAddress) {
```
tinybike: `err` needs to be caught or at least checked for.

scripts/flash/get-timestamp.js
```diff
@@ -0,0 +1,19 @@
...
+    callback({
```
tinybike: Callbacks should use Node.js style: first argument is an error (or `null` if there is no error), second argument contains the results.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var chalk = require("chalk");
+var getTime = require("./get-timestamp");
+const { getPrivateKeyFromString } = require("../dp/lib/get-private-key");
```
tinybike: Don't use `const`, augur.js is (for now) ES5-only.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,103 @@
...
+var displayTime = require("./display-time");
+var setTimestamp = require("./set-timestamp");
+
+const day = 108000; // day
+
+function disputeContributeInternal(augur, marketID, outcome, amount, disputerAuth, invalid, auth, callback) {
+  repFaucet(augur, disputerAuth, function (err) {
+    if (err) { console.log(chalk.red("Error"), chalk.red(err)); callback(err); }
+    if (err) return console.error(err);
+    if (!invalid) { invalid = false; } else { invalid = true; }
```
tinybike: Split this into multiple lines (or use a ternary).

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,103 @@
...
+const repFaucet = require("../rep-faucet");
+var BigNumber = require("bignumber.js");
+var displayTime = require("./display-time");
+var setTimestamp = require("./set-timestamp");
+
+const day = 108000; // day
+
+function disputeContributeInternal(augur, marketID, outcome, amount, disputerAuth, invalid, auth, callback) {
+  repFaucet(augur, disputerAuth, function (err) {
+    if (err) { console.log(chalk.red("Error"), chalk.red(err)); callback(err); }
```
tinybike: Split this into multiple lines.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,103 @@
...
+            displayTime("Current timestamp", timeResult.timestamp);
+            displayTime("Fee Window end time", feeWindowStartTime);
+            displayTime("Set Time to", setTime);
+            setTimestamp(augur, setTime, timeResult.timeAddress, auth, function () {
+              var numTicks = market.numTicks;
+              var payoutNumerators = Array(market.numOutcomes).fill(0);
+              payoutNumerators[outcome] = numTicks;
+              var bnAmount = new BigNumber(amount, 10).toFixed();
+              console.log(chalk.yellow("sending amount REP"), chalk.yellow(bnAmount));
+              augur.api.FeeWindow.isActive(feeWindowPayload, function (err, result) {
```
tinybike: Error needs to be handled.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,103 @@
...
+  repFaucet(augur, disputerAuth, function (err) {
+    if (err) { console.log(chalk.red("Error"), chalk.red(err)); callback(err); }
+    if (err) return console.error(err);
+    if (!invalid) { invalid = false; } else { invalid = true; }
+    augur.markets.getMarketsInfo({ marketIDs: [marketID] }, function (err, marketsInfo) {
+      var market = marketsInfo[0];
+      var marketPayload = { tx: { to: marketID } };
+      augur.api.Market.getFeeWindow(marketPayload, function (err, feeWindowId) {
+        var feeWindowPayload = { tx: { to: feeWindowId } };
+        augur.api.FeeWindow.getStartTime(feeWindowPayload, function (err, feeWindowStartTime) {
```
tinybike: Error needs to be handled.  (This is true for all of these functions, although I'm not going to add a comment for every one.)

scripts/flash/set-timestamp.js
```diff
@@ -0,0 +1,30 @@
...
+      console.log(chalk.green.dim("Success"), chalk.green(JSON.stringify(result)));
+      displayTime("result of time change", timestamp);
+      callback(null);
+    },
+    onFailed: function (result) {
+      console.log(chalk.red.dim("Failed:"), chalk.red(JSON.stringify(result)));
+      callback(null);
```
tinybike: Suggest passing the error to `callback`.

scripts/flash/set-timestamp.js
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var chalk = require("chalk");
+var displayTime = require("./display-time");
+
+function setTimestamp(augur, numberTicks, address, auth, callback) {
```
tinybike: Why is the new timestamp parameter called `numberTicks`?  Should it be called `newTimestamp` instead?

scripts/flash/list-markets.js
```diff
@@ -0,0 +1,46 @@
...
+"use strict";
+
+var chalk = require("chalk");
+var async = require("async");
+
+function listMarketsInternal(augur, universe) {
+  var timestamp = augur.api.Controller.getTimestamp();
+  var currentTime = new Date(timestamp * 1000);
+  augur.markets.getMarkets({ universe: universe, sortBy: "endDate", isSortDescending: true }, function (err, marketIDs) {
+    if (!marketIDs || marketIDs.length === 0) { console.log(chalk.red("No markets available")); process.exit(0);}
```
tinybike: Split into multiple lines.

scripts/flash/designated-report.js
```diff
@@ -0,0 +1,59 @@
...
+    help(callback);
+  } else {
+    var paramArray = params.split(",");
+    var invalid = paramArray.length === 3 ? paramArray[2] : false;
+    var marketID = paramArray[0];
+    var outcomeId = paramArray[1];
+    console.log(chalk.yellow.dim("marketID"), marketID);
+    console.log(chalk.yellow.dim("outcomeId"), outcomeId);
+    console.log(chalk.yellow.dim("invalid"), invalid);
+    designateReportInternal(augur, marketID, outcomeId, invalid, auth, callback);
```
tinybike: `designateReportInternal` is not currently set up to accept a callback.

scripts/flash/designated-report.js
```diff
@@ -0,0 +1,59 @@
...
+      console.log(chalk.red.dim("Market End Time"), chalk.red(endTime));
+      getTime(augur, auth, function (timeResult) {
+        endTime = parseInt(endTime, 10) + 10000;
+        setTimestamp(augur, endTime, timeResult.timeAddress, auth, function () {
+          var numTicks = market.numTicks;
+          var payoutNumerators = Array(market.numOutcomes).fill(0);
+          payoutNumerators[outcomeID] = numTicks;
+
+          doInitialReport(augur, marketID, payoutNumerators, invalid, auth, function (err) {
+            if (err) { console.log(chalk.red(err)); process.exit(1); }
```
tinybike: Split into multiple lines.

scripts/flash/do-initial-report.js
```diff
@@ -0,0 +1,28 @@
...
+    tx: { to: marketID  },
+    _payoutNumerators: payoutNumerators,
+    _invalid: invalid,
+    onSent: function (result) {
+      console.log(chalk.yellow.dim("Sent:"), chalk.yellow(JSON.stringify(result)));
+      console.log(chalk.yellow.dim("Waiting for reply ...."));
+    },
+    onSuccess: function (result) {
+      console.log(chalk.green.dim("Success:"), chalk.green(JSON.stringify(result)));
+      callback(result);
```
tinybike: `result` is passed as the first (and only) argument to the callback, but in scripts/flash/designated-report.js's invocation of this function (line 26) the callback's argument is defined to be an error, and upon receiving an error the script automatically exits.

To avoid this sort of thing, I strongly suggest using Node-style callbacks, with the error always as the first argument, and the result as the second argument.

scripts/flash/finalize-market.js
```diff
@@ -0,0 +1,76 @@
...
+
+"use strict";
+
+var chalk = require("chalk");
+var getTime = require("./get-timestamp");
+
+function finalizeMarketInternal(augur, marketID, auth, callback) {
+  var marketPayload = { tx: { to: marketID } };
+  augur.api.Market.getFeeWindow(marketPayload, function (err, feeWindowAddress) {
+    if (err) { console.error(err); callback(err); }
```
tinybike: Split into multiple lines.

scripts/flash/finalize-market.js
```diff
@@ -0,0 +1,76 @@
...
+    if (feeWindowAddress === "0x0000000000000000000000000000000000000000") {
+      console.log(chalk.red("Market has not be reported on, need initial-report before finalization"));
+      callback(null);
+    }
+    var feeWindowPayload = { tx: { to: feeWindowAddress } };
+    augur.api.FeeWindow.getEndTime(feeWindowPayload, function (err, endTime) {
+      endTime = parseInt(endTime, 10) + 10000; // past fee window end time
+      getTime(augur, auth, function (timeResult) {
+        if (!timeResult) { console.error("Could not get current time"); callback("could not get time"); }
```
tinybike: Split into multiple lines.

scripts/flash/finalize-market.js
```diff
@@ -0,0 +1,76 @@
...
+                console.log(chalk.red.dim("Failed:"), chalk.red(JSON.stringify(result)));
+                process.exit(1);
+              },
+            };
+            console.log(chalk.green.dim("finalizePayload:"), chalk.green(JSON.stringify(finalizePayload)));
+            augur.api.Market.finalize(finalizePayload);
+          },
+          onFailed: function (result) {
+            console.log(chalk.red.dim("Failed:"), chalk.red(JSON.stringify(result)));
+            process.exit(1);
```
tinybike: Suggest only invoking `process.exit` at the top-level.

scripts/flash/get-balance.js
```diff
@@ -0,0 +1,46 @@
...
+  callback(null);
+}
+
+function getBalance(augur, params, callback) {
+  if (!params || params === "help") {
+    help(callback);
+  } else {
+    var account = params;
+    var universe = augur.contracts.addresses[augur.rpc.getNetworkID()].Universe;
+    getBalanceInternal(augur, universe, account, callback);
```
tinybike: `getBalanceInternal` does not accept a callback.

scripts/flash/get-balance.js
```diff
@@ -0,0 +1,46 @@
...
+var speedomatic = require("speedomatic");
+
+
+function getBalanceInternal(augur, universe, account) {
+  console.log(chalk.green.dim("address:"), chalk.green(account));
+  console.log(chalk.green.dim("universe:"), chalk.green(universe));
+  augur.api.Universe.getReputationToken({ tx: { to: universe } }, function (err, reputationTokenAddress) {
+    if (err) { console.error("Could not get universe"); process.exit(1); }
+    augur.api.ReputationToken.balanceOf({ tx: { to: reputationTokenAddress }, _owner: account }, function (err, reputationBalance) {
+      if (err) { console.error("ReputationToken.balanceOf failed:", err); process.exit(1); }
```
tinybike: Split into multiple lines.

scripts/flash/get-balance.js
```diff
@@ -0,0 +1,46 @@
...
+
+var chalk = require("chalk");
+var speedomatic = require("speedomatic");
+
+
+function getBalanceInternal(augur, universe, account) {
+  console.log(chalk.green.dim("address:"), chalk.green(account));
+  console.log(chalk.green.dim("universe:"), chalk.green(universe));
+  augur.api.Universe.getReputationToken({ tx: { to: universe } }, function (err, reputationTokenAddress) {
+    if (err) { console.error("Could not get universe"); process.exit(1); }
```
tinybike: Split into multiple lines.

scripts/flash/index.js
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env node
```
tinybike: This whole file needs to be converted to ES5 syntax.

scripts/flash/index.js
```diff
@@ -0,0 +1,156 @@
...
+    const connect = promisify(augur.connect);
+    const auth = getPrivateKeyFromString(network.privateKey);
+
+    await connect({ ethereumNode: { http: network.http }, augurNode: process.env.AUGUR_WS });
+    await methods[commands.indexOf(command)](augur, params, auth);
+
+  }
+}
+
+function parseArgs() {
```
tinybike: Suggest using a standard parameter parser like `getopt` instead of rolling your own, as the custom syntax is (to me) somewhat confusing and unclear.

scripts/flash/initial-report.js
```diff
@@ -0,0 +1,76 @@
...
+          setTimestamp(augur, endTime, timeResult.timeAddress, auth, function () {
+            var numTicks = market.numTicks;
+            var payoutNumerators = Array(market.numOutcomes).fill(0);
+            payoutNumerators[outcome] = numTicks;
+            doInitialReport(augur, marketID, payoutNumerators, invalid, userAuth, function (err) {
+              if (err) {
+                console.log(chalk.red(err));
+                process.exit(1);
+              }
+              console.log(chalk.green("Initial Report Done"));
```
tinybike: Should `callback` be called here?

scripts/flash/push-time.js
```diff
@@ -0,0 +1,65 @@
...
+function pushTime(augur, params, auth, callback) {
+  var regex = /\d+/g;
+  if (!params || params === "help" || params.split(",").length < 2) {
+    help(callback);
+  } else {
+    var paramArray = params.split(",");
+    var action = paramArray[0];
+    var value = paramArray[1];
+    getTime(augur, auth, function (timeResult) {
+      var timestamp = parseInt(augur.api.Controller.getTimestamp(), 10);
```
tinybike: Needs callback for async.

scripts/flash/list-markets.js
```diff
@@ -0,0 +1,46 @@
...
+  callback(null);
+}
+
+function listMarkets(augur, params, callback) {
+  if (params === "help") {
+    help(callback);
+  } else {
+    var universe = augur.contracts.addresses[augur.rpc.getNetworkID()].Universe;
+    console.log(chalk.green.dim("Universe"), universe);
+    listMarketsInternal(augur, universe, callback);
```
tinybike: `listMarketsInternal` does not accept a callback.

scripts/flash/list-markets.js
```diff
@@ -0,0 +1,46 @@
...
+
+function listMarketsInternal(augur, universe) {
+  var timestamp = augur.api.Controller.getTimestamp();
+  var currentTime = new Date(timestamp * 1000);
+  augur.markets.getMarkets({ universe: universe, sortBy: "endDate", isSortDescending: true }, function (err, marketIDs) {
+    if (!marketIDs || marketIDs.length === 0) { console.log(chalk.red("No markets available")); process.exit(0);}
+    augur.markets.getMarketsInfo({ marketIDs: marketIDs }, function (err, marketInfos) {
+      if (!marketInfos || !Array.isArray(marketInfos) || !marketInfos.length) return;
+      var infos = marketInfos.sort(function (a, b) { return b.endDate - a.endDate; });
+      async.eachSeries(infos, function (marketInfo, nextMarket) {
```
tinybike: Suggest using an ordinary for loop (or `forEach` etc.) instead of `async.eachSeries`, since nothing in the body of the loop is asynchronous.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,106 @@
...
+              console.log(chalk.yellow("sending amount REP"), chalk.yellow(bnAmount));
+              augur.api.FeeWindow.isActive(feeWindowPayload, function (err, result) {
+                console.log(chalk.green.dim("Few Window is active"), chalk.green(result));
+                if (result) {
+                  augur.api.Market.contribute({
+                    meta: disputerAuth,
+                    tx: { to: marketID  },
+                    _payoutNumerators: payoutNumerators,
+                    _invalid: invalid,
+                    _amount: bnAmount,
```
tinybike: Numerical arguments to the autogenerated api (`augur.api.*`) should always be provided as 0x-prefixed hex strings; `bnAmount` is a base-10 string.  (Regular JS numbers will work too, but I suggest avoiding these for various reasons.)  Suggest using `"0x" + new BigNumber(amount, 10).toString(16)` on line 38 to obtain a hex string.

scripts/flash/dispute-contribute.js
```diff
@@ -0,0 +1,106 @@
...
+                      console.log(chalk.green.dim("Success Dispute:"), chalk.green(JSON.stringify(result)));
+                      callback(null);
+                    },
+                    onFailed: function (result) {
+                      console.log(chalk.red.dim("Failed Dispute:"), chalk.red(JSON.stringify(result)));
+                      callback(result);
+                    },
+                  });
+                } else {
+                  console.log(chalk.red("Fee Window isn't active"));
```
tinybike: Should `callback` be called here?

scripts/flash/finalize-market.js
```diff
@@ -0,0 +1,76 @@
...
+var chalk = require("chalk");
+var getTime = require("./get-timestamp");
+
+function finalizeMarketInternal(augur, marketID, auth, callback) {
+  var marketPayload = { tx: { to: marketID } };
+  augur.api.Market.getFeeWindow(marketPayload, function (err, feeWindowAddress) {
+    if (err) { console.error(err); callback(err); }
+    if (feeWindowAddress === "0x0000000000000000000000000000000000000000") {
+      console.log(chalk.red("Market has not be reported on, need initial-report before finalization"));
+      callback(null);
```
tinybike: Do you intend for the function to continue executing after `callback` is called?  If not, should use `return callback(null);`.

scripts/flash/get-balance.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+  });
+}
+
+function help(callback) {
+  console.log(chalk.red("account address is needed to get balances"));
+  callback(null);
+}
+
+function getBalance(augur, params, callback) {
```
tinybike: Suggest using the `getBalances` function in `dp/lib` instead of rolling your own (nearly?) identical function here.

src/contracts/addresses.json
```diff
@@ -44,21 +44,21 @@
   "TradingEscapeHatch": "0x2709bbb7409fa4a336f151ef8b94cf5d15e1911b"
  },
  "12346": {
```
bthaile: I don't think we are suppose to check in this file. @pgebheim can confirm.

scripts/flash/get-balance.js
bthaile: changed to use dp/lib/get-balances

>scripts/flash -c get-balances -p <user address> -n environment

is the execution format, changed to use options parser as suggested. This removes the `params=` and `networks=` in command line. I'd like to sync up with @pgebheim on format so that dp and flash have same command-line feel.

scripts/flash/get-balance.js
bthaile: updated to take callback

scripts/flash/index.js
bthaile: updated file to be ES5 format

scripts/flash/list-markets.js
bthaile: updated to take callback.

scripts/flash/dispute-contribute.js
bthaile: added error handling, tried to find them all in all scripts.

scripts/flash/finalize-market.js
bthaile: did split into multiple lines for clarity.

scripts/flash/get-timestamp.js
bthaile: converted the sync to async calls to get Controller getTimestamp()

scripts/flash/index.js
bthaile: now using options parser node package

scripts/flash/list-markets.js
bthaile: split up the lines

scripts/flash/set-timestamp.js
bthaile: update the parameter

scripts/flash/dispute-contribute.js
bthaile: This was changed, don't know why the change isn't reflected here. should be

                 } else {
                  console.log(chalk.red("Fee Window isn't active"));
                  callback("Fee Window isn't active");

PR #240, opened 2018-02-15
add time constants

epheph:
these values are part of processes where states change without transactions, so we need to know them to know when advanceTime logic is appropriate.


epheph:
Created from:

https://github.com/AugurProject/augur-core/blob/master/source/contracts/reporting/Reporting.sol#L5

PR #241, opened 2018-02-19
Create order book depth around random market midpoint

stephensprinkle:


bthaile:
Looks good, we need a full order book

PR #244, opened 2018-02-21
remove special case handling for event being used in augur node

nuevoalex:
Winning Tokens redeemed is no longer an event and the tokens transferred values being a different format than other token events is causing issues in augur node.


tinybike:
Linter failing though, fix that before merging

PR #247, opened 2018-02-21
options parser got missed on merge

bthaile:
Need options parser for flash scripts to work, looks like options parser got missed on merge


bthaile:
weird travis error, rebasing and creating new branch.

PR #256, opened 2018-02-27
Add `numTicks` param to `placeTrade` function

adrake33:
`augur.trading.placeTrade` was throwing the following exception when I tried calling it via the JS console:

```
Uncaught Error: new BigNumber() not a number: undefined
    at raise (bignumber.js?8748:1181)
    at eval (bignumber.js?8748:1169)
    at new BigNumber (bignumber.js?8748:195)
    at new BigNumber (bignumber.js?8748:205)
    at convertDecimalToFixedPoint (convert-decimal-to-fixed-point.js?8055:14)
    at calculateTradeCost (calculate-trade-cost.js?7a38:17)
    at tradeUntilAmountIsZero (trade-until-amount-is-zero.js?a628:33)
    at eval (place-trade.js?466a:39)
    at dispatchJsonRpcResponse (dispatch-json-rpc-response.js?4091:32)
    at WebSocket.webSocketClient.onmessage (ws-transport.js?9356:26)
```

This was happening because the function `tradeUntilAmountIsZero` was looking for `p.numTicks`, which was undefined.

### DIFF
--- a/src/trading/place-trade.js
+++ b/src/trading/place-trade.js
@@ -13,6 +13,7 @@ var normalizePrice = require("./normalize-price");
  * @param {string} p.estimatedCost Total cost (in ETH) of this trade, as a base-10 string.
  * @param {string} p.minPrice The minimum display (non-normalized) price for this market, as a base-10 string.
  * @param {string} p.maxPrice The maximum display (non-normalized) price for this market, as a base-10 string.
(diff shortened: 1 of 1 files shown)


tinybike:
numTicks should be automatically passed along without needing to specify it -- it's part of p.  I don't think this is necessary.

adrake33:
Merging since this was actually just a missing param in the comments.

PR #261, opened 2018-03-02
Flash help fix

bthaile:
help text fixes
added lots of gas for initial report and dispute contribute


bthaile:
updated to use constant TRADE_GAS

PR #262, opened 2018-03-02
2576 audit cleanup

tinybike:
Tagging @adrake33 for a consistency check across many repos :grin:


scripts/helpers/check-mailbox.js
```diff
@@ -15,12 +15,13 @@ var augur = new Augur();
...
-    if (err) return console.log(err);
+    if (err) return console.error(err);
     console.log(chalk.green.dim("market mailbox address:"), chalk.green(marketMailboxAddress));
     augur.api.Cash.getBalance({ _address: marketMailboxAddress }, function (err, cashBal) {
-      if (err) return console.log(err);
+      if (err) return console.error(err);
       var cashBalance = speedomatic.bignum(cashBal);
-      augur.rpc.eth.getBalance([marketMailboxAddress, "latest"], function (attoEthBalance) {
+      augur.rpc.eth.getBalance([marketMailboxAddress, "latest"], function (err, attoEthBalance) {
+        if (err) return console.error(err);
```
adrake33: Should add check like: 
```
if (attoEthBalance == null) return callback(new Error("rpc.eth.getBalance failed"));
```

src/api/bind-contract-function.js
```diff
@@ -23,9 +23,8 @@ function bindContractFunction(functionAbi) {
       }
       if (isFunction(params[params.length - 1])) callback = params.pop();
       if (!isFunction(callback)) return ethrpc.callContractFunction(payload);
-      return ethrpc.callContractFunction(payload, function (response) {
-        if (response == null) return callback("No response");
-        if (response.error != null) return callback(response.error);
+      return ethrpc.callContractFunction(payload, function (err, response) {
+        if (err) return callback(err);
```
adrake33: Should this also have a check to ensure `response` is not null?

tinybike: Hm, no I don't think that is necessary -- if response is null, ethrpc should pass back an error.

scripts/helpers/check-mailbox.js
tinybike: Hm...the only thing that was updated in this PR was the callback signature (and changing console.log to console.error).  I think if the result that comes back is `null`, ethrpc will pass back an error object for `err`.

src/api/bind-contract-function.js
tinybike: Actually it appears you're right -- ethrpc is just checking for `undefined`, not `null` results.  So there is an extra check needed here.  Nice catch!

PR #268, opened 2018-03-08
more references we need to get rid of

epheph:
These being here cause augur.js to fail when it is included (although it doesn't fail an tests!)

### DIFF
--- a/src/markets/index.js
+++ b/src/markets/index.js
@@ -2,11 +2,8 @@
 
 module.exports = {
   getMarketsInfo: require("./get-markets-info"),
-  getDisputableMarkets: require("./get-disputable-markets"),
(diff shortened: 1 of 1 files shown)


epheph:
here's what i was getting

```
> augur-node@0.1.0 start /Users/epheph/augur/augur-node
> node ./build

module.js:529
    throw err;
    ^

Error: Cannot find module './get-markets-awaiting-reporting'
    at Function.Module._resolveFilename (module.js:527:15)
    at Function.Module._load (module.js:476:23)
    at Module.require (module.js:568:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/epheph/augur/augur.js/src/markets/index.js:7:32)
```

PR #269, opened 2018-03-09
5524 merge ethereumjs connect into augurjs

tinybike:
Resolves [ch5524] https://app.clubhouse.io/augur/story/5524/merge-ethereumjs-connect-into-augur-js

Decided to go ahead and knock this out, because having ethereumjs-connect as a separate, unnecessary repo is a (minor) headache that's caused a few hard-to-debug problems. Plus remembering to push updates to this repo is a pain. Now we won't have to, yay.


src/connect/connect-to-ethereum.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
```
adrake33: Double check

src/connect/index.js
```diff
@@ -1,17 +1,17 @@
 "use strict";
 
 var async = require("async");
-var ethereumConnector = require("ethereumjs-connect");
```
adrake33: It looks like ethereumjs-connect is still used in the augur-node and augur repos. I assume the plan to remove it from those as well once this is merged in?

src/connect/create-ethrpc-configuration.js
```diff
@@ -0,0 +1,16 @@
...
+
+function createEthrpcConfiguration(configuration) {
+  var ethrpcConfiguration = {
+    connectionTimeout: 60000,
+    errorHandler: function (err) { if (err) console.error(err); },
+  };
+  ethrpcConfiguration.httpAddresses = configuration.httpAddresses;
+  ethrpcConfiguration.wsAddresses = configuration.wsAddresses;
+  ethrpcConfiguration.ipcAddresses = configuration.ipcAddresses;
+  ethrpcConfiguration.networkID = configuration.networkID;
```
adrake33: I think this needs to match the casing used on line 51 of src/connect/index.js (i.e. it should be `networkId`, not `networkID`): https://github.com/AugurProject/augur.js/blob/52e5c330914adb2813b2093f1a694e6772c84533/src/connect/index.js#L51

src/connect/set-contracts.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+function setContracts(networkID, allContracts) {
```
adrake33: Actually, this (and other such instances) should also be changed to `networkId` to follow the convention used elsewhere.

adrake33: Yeah, I realize it's not possible to update every instance. Might as well update as many as we can, though.

PR #277, opened 2018-03-12
pulled out calling rep faucet

bthaile:
since the creation of fork scripts and changes to rep-faucet. It was missed that dispute-contribute needs to pass in amount of rep. 
Changed to pull out calling rep-faucet so that we call rep-faucet only if we need rep.

### DIFF
--- a/scripts/flash/designated-report.js
+++ b/scripts/flash/designated-report.js
@@ -5,6 +5,7 @@
 var chalk = require("chalk");
 var getTime = require("./get-timestamp");
 var setTimestamp = require("./set-timestamp");
+var getRepTokens = require("./get-rep-tokens");
 var doInitialReport = require("./do-initial-report");
 var getPayoutNumerators = require("./get-payout-numerators");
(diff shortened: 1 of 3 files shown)


scripts/flash/designated-report.js
```diff
@@ -13,34 +14,41 @@ var getPrivateKeyFromString = require("../dp/lib/get-private-key").getPrivateKey
  * Move time to Market end time and do initial report
  */
 function designateReportInternal(augur, marketId, outcome, userAuth, invalid, auth, callback) {
-  augur.markets.getMarketsInfo({ marketIds: [marketId] }, function (err, marketsInfo) {
-    var market = marketsInfo[0];
-    var marketPayload = { tx: { to: marketId } };
-    // make sure we have the correct account to report with
-    if (userAuth.address !== market.designatedReporter && auth.address === market.designatedReporter) {
-      userAuth = auth;
+  var amount = 10000;
```
justinbarry: @bthaile Should this value be hardcoded?

bthaile: hardcoding is fine for this application, nice to have a high default value.

PR #280, opened 2018-03-13
7852 scalar price fixes

tinybike:
Resolves:
https://app.clubhouse.io/augur/story/7852/scalar-order-creation-transactions-fail
https://app.clubhouse.io/augur/story/7863/speed-up-canned-markets-scripts-called-from-dp
https://app.clubhouse.io/augur/story/7864/make-polling-block-interval-an-adjustable-parameter-in-augur-js


test/unit/trading/calculate-trade-cost.js
```diff
@@ -0,0 +1,110 @@
...
+describe("trading/calculate-trade-cost", function () {
+  var test = function (t) {
+    it(t.description, function () {
+      t.assertions(calculateTradeCost(t.params));
+    });
+  };
+  test({
+    description: "scalar bid",
+    params: {
+      displayPrice: "20",
```
MicahZoltu: Recommend one or more test cases with a negative `displayPrice`.

scripts/set-local-dev-env.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # All of these keys are optional, and have the defaults listed here
-export USE_NORMAL_TIME="false"                # need to beable to change time manually
+# export USE_NORMAL_TIME="false"                # need to beable to change time manually
```
nuevoalex: why remove this?

tinybike: I forgot to uncomment it D:
Will fix

test/unit/trading/calculate-trade-cost.js
tinybike: Good thinking!  Do you want to write a couple?  (You know you do)

PR #284, opened 2018-03-15
extra info signature change

epheph:
I modified the 3 create market functions to accept extraInfo whitelist'd keys either in the _extraInfo object OR just attached to the top level.

### DIFF
--- a/src/create-market/split-extra-info.js
+++ b/src/create-market/split-extra-info.js
@@ -1,0 +1,18 @@
+"use strict";
+
+var assign = require("lodash.assign");
+var immutableDelete = require("immutable-delete");
+
+function splitExtraInfo(p) {
+  var extraInfoKeys = ["longDescription", "tags", "minPrice", "maxPrice", "_scalarDenomination"];
+  var extraInfo = extraInfoKeys.reduce(function (prev, extraInfoKey) {
+    if (p.hasOwnProperty(extraInfoKey) && !prev.hasOwnProperty(extraInfoKey)) prev[extraInfoKey] = p[extraInfoKey];
+    return prev;
+  }, assign({}, p._extraInfo));
+  return assign({},
+    immutableDelete(p, extraInfoKeys),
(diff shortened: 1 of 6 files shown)


src/create-market/split-extra-info.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+var assign = require("lodash.assign");
+var immutableDelete = require("immutable-delete");
+
+function splitExtraInfo(p) {
+  var extraInfoKeys = ["longDescription", "tags", "_scalarDenomination", "resolutionSource", "outcomeNames"];
+  var extraInfo = extraInfoKeys.reduce(function (prev, extraInfoKey) {
```
bthaile: I think knowing the extra info names makes this too  tightly bound.

epheph: @stephensprinkle was asking for this, so he can just submit the data via a function call and let the middleware worry about the implementation detail of which data structure it belongs in. I could move it up into a constant, but it's either in the UI or in augur.js

bthaile: is this change needed for the UI to get scalar denomination in market-info?

epheph: No, this just allows you to call createMarket as:
```
{longDescription: ""}
```

OR
```
{ extraInfo: { longDescription: "" } }
```

tinybike: While it seems reasonable for the UI to not concern itself with whether fields are part of `extraInfo`, IMO this is not something we should worry about for MVP...  However, a detail that the UI needs to be aware of is whether a text field is actually `bytes32` (`_topic` and each element of `_outcomes` -- these can be 32 bytes max), or if it's `string` (`_extraInfo` and `_description`).  If the UI needs to know these lower-level details, I don't really see the point of hiding what goes in extraInfo from the UI.

tinybike:
Closing this for now.

PR #286, opened 2018-03-18
8045 fill price fix

tinybike:
Resolves [ch8045](https://app.clubhouse.io/augur/story/8045/number-of-shares-remaining-not-calculated-correctly-when-filling-bid-orders) and [ch8046](https://app.clubhouse.io/augur/story/8046/decrease-default-trade-gas-to-6-175m)


scripts/dp/lib/create-order.js
```diff
@@ -85,16 +85,10 @@ function createOrder(augur, marketId, outcome, numOutcomes, maxPrice, minPrice,
...
-        if (debugOptions.cannedMarkets) console.log(chalk.green.dim("placeTrade success:"), tradeOnChainAmountRemaining);
-        callback(null);
-      },
-      onFailed: function (err) {
-        if (debugOptions.cannedMarkets) console.error(chalk.red.bold("placeTrade failed:"), err);
-        callback(err);
+        if (debugOptions.cannedMarkets) console.log(chalk.green.dim("placeTrade sent:"), res.hash);
       },
+      onSuccess: function () { callback(null); },
+      onFailed: callback,
```
adrake33: It seems like this should be passing an error back to the callback function. Why was this removed?

tinybike: It still does.

PR #295, opened 2018-03-23
added approval flash script.

bthaile:
simple script to approve an account.  pass in user address, if not then use existing env. variable for private key. I thought this would be easier to approve multiple accounts via a command line.

### DIFF
--- a/scripts/flash/approve-account.js
+++ b/scripts/flash/approve-account.js
@@ -28,7 +28,7 @@ function approveAccount(augur, params, auth, callback) {
     help(callback);
   } else {
     var account = params;
-    if (account === undefined) {
+    if (account == null) {
       account = auth.address;
     }
     var universe = augur.contracts.addresses[augur.rpc.getNetworkID()].Universe;
--- a/scripts/flash/approve-account.js
+++ b/scripts/flash/approve-account.js
@@ -1,0 +1,39 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var chalk = require("chalk");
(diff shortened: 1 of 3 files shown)


scripts/flash/approve-account.js
```diff
@@ -0,0 +1,39 @@
...
+  approveAugurEternalApprovalValue(augur, address, auth, function (err) {
+    if (err) {
+      console.log(chalk.red(err));
+      return callback(JSON.stringify(err));
+    }
+    callback(null);
+  });
+}
+
+function help(callback) {
```
tinybike: Why does `help` require a `callback` argument?  It does not perform any asynchronous operations.

scripts/flash/approve-account.js
```diff
@@ -0,0 +1,39 @@
...
+  console.log(chalk.red("account address is needed to get approval"));
+  callback(null);
+}
+
+function approveAccount(augur, params, auth, callback) {
+  if (params === "help") {
+    help(callback);
+  } else {
+    var account = params;
+    if (account === undefined) {
```
tinybike: Suggest `if (account == null) {` instead to handle both `null` and `undefined` inputs.

scripts/flash/approve-account.js
bthaile: the call to the script is passed to help() a callback needs to be called to tell index.js the script is done running. index.js doesn't know help is called it's just executing the script.

scripts/flash/approve-account.js
bthaile: right, forgot that one. more convenient. Will make change.

PR #296, opened 2018-03-26
Docker copy addresses

bthaile:
added npm command to copy addresses from docker pop latest.


epheph:
Add upload block numbers, too?

adrake33:
Looks fine to me, but as Scott mentioned, adding upload block number would be good.

PR #297, opened 2018-03-26
added list-market-orders 

bthaile:
added simple script to list out orders on market. quick way to see order info. added script to fill all orders based on buy | sell and outcomeId. Mostly thinking about binary and category markets.

### DIFF
--- a/scripts/flash/fill-market-orders.js
+++ b/scripts/flash/fill-market-orders.js
@@ -1,0 +1,81 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var async = require("async");
+var chalk = require("chalk");
+var fillOrder = require("../dp/lib/fill-order");
+
+function fillMarketOrderInternal(augur, marketIds, orderType, outcomeId, universe, auth, callback) {
+  augur.markets.getMarkets({ universe: universe, sortBy: "endDate", isSortDescending: true }, function (err, marketIds) {
+    if (!marketIds || marketIds.length === 0) {
+      console.log(chalk.red("No markets available"));
+      callback("No Markets");
+    }
+    augur.markets.getMarketsInfo({ marketIds: marketIds }, function (err, marketInfos) {
+      if (err) {
+        console.log(chalk.red("Error "), chalk.red(err));
+        return callback(err);
+      }
+      if (!marketInfos || !Array.isArray(marketInfos) || !marketInfos.length) {
+        return callback("No Market Info");
+      }
+      var market = marketInfos[0];
+      var marketId = market.id;
+      console.log(chalk.yellow.dim("outcome:"), chalk.yellow(outcomeId), chalk.yellow.dim("order type:"), chalk.yellow(orderType));
+      console.log(chalk.green.dim("OrderId:"), chalk.green.dim("Account:"), chalk.green.dim("Amount:"), chalk.green.dim("FullPrecisionPrice:"), chalk.green.dim("FullPrecisionAmount"));
+      augur.trading.getOrders({ marketId: marketId, outcome: outcomeId, orderType: orderType }, function (err, orderBook) {
+        if (err) {
+          console.error(err);
+          callback(err);
+        }
+        if (!orderBook[marketId]) {
+          callback("No Market Orders Found");
+        }
+        var orders = orderBook[marketId][outcomeId][orderType];
+        async.eachSeries(Object.keys(orders), function (orderId, nextOrder) {
+          var order = orders[orderId];
+          console.log(chalk.yellow(order.fullPrecisionPrice), chalk.yellow(order.fullPrecisionAmount));
+          fillOrder(augur, universe, auth.address, outcomeId, order.fullPrecisionAmount, orderType, auth, function (err) {
+            if (err) {
+              nextOrder(err);
+            }
+            console.log(chalk.green("Success"));
+            nextOrder(null);
+          });
+        }, function (err) {
+          callback(err);
+        });
+      });
+    });
+  });
+
+}
+
+function help(callback) {
+  console.log(chalk.red("params syntax --> marketId"));
+  console.log(chalk.red("params syntax --> outcome ( 0, 1, ... )"));
+  console.log(chalk.red("params syntax --> order type (buy | sell)"));
+  console.log(chalk.yellow("script will approve account"));
+  callback(null);
+}
+
+function fillMarketOrder(augur, params, auth, callback) {
+  console.log("params", params);
+  if (params === "help") {
+    help(callback);
+  } else {
+    if (params == null) return callback("MarketId is required");
+    var paramArray = params.split(",");
+    var marketId = paramArray[0];
+    var outcomeId = paramArray[1];
+    var orderType = paramArray[2];
+
+    console.log(chalk.yellow.dim("marketId"), chalk.yellow(marketId));
+    var universe = augur.contracts.addresses[augur.rpc.getNetworkID()].Universe;
+    console.log(chalk.yellow.dim("Universe"), chalk.yellow(universe));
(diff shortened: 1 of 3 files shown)


scripts/flash/fill-market-orders.js
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var async = require("async");
+var chalk = require("chalk");
+var fillOrder = require("../dp/lib/fill-order");
+
+function fillMarketOrderInternal(augur, marketIds, orderType, outcomeId, universe, auth, callback) {
```
epheph: This `marketIds` parameter doesn't seem to be used, it is immediately overwritten by a callback

scripts/flash/fill-market-orders.js
```diff
@@ -0,0 +1,81 @@
...
+  augur.markets.getMarkets({ universe: universe, sortBy: "endDate", isSortDescending: true }, function (err, marketIds) {
+    if (!marketIds || marketIds.length === 0) {
```
epheph: `err` check

scripts/flash/fill-market-orders.js
```diff
@@ -0,0 +1,81 @@
...
+        if (!orderBook[marketId]) {
+          callback("No Market Orders Found");
+        }
+        var orders = orderBook[marketId][outcomeId][orderType];
+        async.eachSeries(Object.keys(orders), function (orderId, nextOrder) {
+          var order = orders[orderId];
+          console.log(chalk.yellow(order.fullPrecisionPrice), chalk.yellow(order.fullPrecisionAmount));
+          fillOrder(augur, universe, auth.address, outcomeId, order.fullPrecisionAmount, orderType, auth, function (err) {
+            if (err) {
+              nextOrder(err);
```
epheph: You should `return nextError(err)` here, or else you'll call `nextOrder` again below

scripts/flash/fill-market-orders.js
```diff
@@ -0,0 +1,81 @@
...
+        return callback("No Market Info");
+      }
+      var market = marketInfos[0];
+      var marketId = market.id;
+      console.log(chalk.yellow.dim("outcome:"), chalk.yellow(outcomeId), chalk.yellow.dim("order type:"), chalk.yellow(orderType));
+      console.log(chalk.green.dim("OrderId:"), chalk.green.dim("Account:"), chalk.green.dim("Amount:"), chalk.green.dim("FullPrecisionPrice:"), chalk.green.dim("FullPrecisionAmount"));
+      augur.trading.getOrders({ marketId: marketId, outcome: outcomeId, orderType: orderType }, function (err, orderBook) {
+        if (err) {
+          console.error(err);
+          callback(err);
```
epheph: More `return`s required around callback(err) in a few other places

scripts/flash/fill-market-orders.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    augur.markets.getMarketsInfo({ marketIds: marketIds }, function (err, marketInfos) {
+      if (err) {
+        console.log(chalk.red("Error "), chalk.red(err));
+        return callback(err);
+      }
+      if (!marketInfos || !Array.isArray(marketInfos) || !marketInfos.length) {
+        return callback("No Market Info");
+      }
+      var market = marketInfos[0];
```
epheph: Only does the first one? Not sure what the expected behavior for this script is

scripts/flash/list-market-orders.js
```diff
@@ -0,0 +1,84 @@
...
+          return callback(err);
+        }
+        if (!marketInfos || !Array.isArray(marketInfos) || !marketInfos.length) {
+          return callback("No Market Info");
+        }
+        var market = marketInfos[0];
+        var marketId = market.id;
+        var outcomes = [0, 1];
+        if (market.marketType === "category") {
+          outcomes = new Array(market.numOutcomes);
```
epheph: `market.numOutcomes` will always have the right number, you could just do this one line every time instead of switching when it is `category`

scripts/flash/index.js
```diff
@@ -19,10 +19,12 @@ var marketInfo = require("./market-info");
 var showInitialReporter = require("./show-initial-reporter");
 var fork = require("./fork");
 var approval = require("./approve-account");
+var listMarketOrders = require("./list-market-orders");
+var fillMarketOrders = require("./fill-market-orders");
 
-var commands = ["get-balance", "list-markets", "designate-report", "initial-report", "dispute-contribute", "finalize-market", "push-time", "market-info", "show-initial-reporter", "fork", "approval"];
+var commands = ["get-balance", "list-markets", "designate-report", "initial-report", "dispute-contribute", "finalize-market", "push-time", "market-info", "show-initial-reporter", "fork", "approval", "list-market-orders", "fill-market-orders"];
 var NETWORKS = ["aura", "clique", "environment", "rinkeby", "ropsten"];
-var methods = [getBalance, listMarkets, designatedReport, initialReport, disputeContribute, finalizeMarket, pushTime, marketInfo, showInitialReporter, fork, approval];
```
epheph: These two lists are closely related, given how large they have become, it's hard to read. Might be worth converting it to an object which maps command to method (or if they could be the exact same thing, that would be great, too)

adrake33:
Don't have much else to add that Scott hasn't already said. 

It seems like `marketIds` on line 9 of scripts/flash/fill-market-orders.js should be `marketId`, and line 23 should be getting the market info for that particular `marketId`. Otherwise, `fillMarketOrder`and `fillMarketOrderInternal` should be renamed to indicate multiple markets can be passed in.

scripts/flash/fill-market-orders.js
bthaile: Only takes in one market. No endpoint that takes a single marketId. See the fillMarketOrderIneternal call.

scripts/flash/fill-market-orders.js
bthaile: Copy paste typo. I’ll change to marketId. I’ll make the chamge for clarity

scripts/flash/fill-market-orders.js
epheph: OK, makes sense. So i think the issue is at the top of this function, the passed in market list isn't being used at all, and it just grabs all markets sorted by endDate.

scripts/flash/fill-market-orders.js
bthaile: copy paste issue, I'll fix.

scripts/flash/fill-market-orders.js
bthaile: I'll clean up the marketId by removing the initial getMarkets(...) call, it's not needed, another copy paste typo 🤕

bthaile:
Fixed up index.js, good call!
cleaned up typos, should be gtg.

scripts/flash/index.js
```diff
@@ -19,35 +19,92 @@ var marketInfo = require("./market-info");
 var showInitialReporter = require("./show-initial-reporter");
 var fork = require("./fork");
 var approval = require("./approve-account");
+var listMarketOrders = require("./list-market-orders");
+var fillMarketOrders = require("./fill-market-orders");
 
-var commands = ["get-balance", "list-markets", "designate-report", "initial-report", "dispute-contribute", "finalize-market", "push-time", "market-info", "show-initial-reporter", "fork", "approval"];
 var NETWORKS = ["aura", "clique", "environment", "rinkeby", "ropsten"];
-var methods = [getBalance, listMarkets, designatedReport, initialReport, disputeContribute, finalizeMarket, pushTime, marketInfo, showInitialReporter, fork, approval];
+var methods = [
```
epheph: Much easier to read! I think it could be even more improved if the key was the command and the value was the method, but this is def an improvement

epheph:
nice, ship it! 👍

PR #304, opened 2018-03-28
Build geth instance with contracts and normal time as well as timetravel

pgebheim:


scripts/push-docker.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
```
bthaile: Are you concerned that augur-core tag could be changed between building and pushing?

if they build then next day update package.json and then do a push

package.json
```diff
@@ -20,7 +20,9 @@
     "preversion": "npm test",
     "version": "npm run build && git add -A dist src/version.js package.json package-lock.json",
     "dp": "node scripts/dp",
-    "docker:build-and-push": "scripts/build-docker.sh",
+    "docker:build": "scripts/build-docker.sh",
+    "docker:push": "scripts/push-docker.sh",
+    "docker:build-and-push": "npm run docker:build && npm run docker:push",
```
bthaile: recommend removing push only, just to make sure they push is recent build.

package.json
```diff
@@ -43,9 +42,12 @@
...
     "release:major": "npm version major && git push && git push --tags && npm publish",
+    "docker:build": "scripts/build-docker.sh",
+    "docker:build-and-push": "npm run docker:build && scripts/push-docker.sh",
     "docker:pull": "docker pull augurproject/dev-node-parity:latest && docker pull augurproject/dev-node-geth:latest && docker pull augurproject/dev-pop-geth:latest",
     "docker:geth": "docker run -e GETH_VERBOSITY=4 -it -p 8545:8545 -p 8546:8546 -v $HOME/docker:/geth/chain augurproject/dev-node-geth:latest",
     "docker:geth:pop": "docker run -e GETH_VERBOSITY=4 -it -p 8545:8545 -p 8546:8546 augurproject/dev-pop-geth:latest",
+    "docker:geth:pop-normal-time": "docker run -e GETH_VERBOSITY=4 -it -p 8545:8545 -p 8546:8546 augurproject/dev-pop-normtime-geth:latest",
     "docker:geth:attach": "docker run --rm --net host -it ethereum/client-go attach rpc:http://127.0.0.1:8545",
     "docker:geth:files": "scripts/copy-docker-files.sh",
     "docker:geth:build": "scripts/build-docker.sh"
```
bthaile: this is same as docker:build

probably just need one.

PR #309, opened 2018-03-29
pass in a shorter polling interval

bthaile:
shorter polling interval.


bthaile:
takes forcing a fork from 5 minutes or so to less than 2 or so.

PR #312, opened 2018-03-30
Change how claimReportingFees passes errors to callback

adrake33:
All other Simplified API functions pass an error param to the callback, but since this function executes numerous transactions and returns an array of which ones failed, this PR modifies that behavior to pass a generic error to the callback.  It also renames the `errors` property to `failedTransactions` and slightly modifies the shape of its data.

### DIFF
--- a/src/reporting/claim-reporting-fees.js
+++ b/src/reporting/claim-reporting-fees.js
@@ -15,7 +15,7 @@ var PARALLEL_LIMIT = require("../constants").PARALLEL_LIMIT;
  */
 
  /**
- * @typedef {Object} gasEstimateTotals
+ * @typedef {Object} GasEstimateTotals
  * @property {BigNumber} disputeCrowdsourcers Total gas estimate for redeeming all DisputeCrowdsourcer reporting fees.
  * @property {BigNumber} initialReporters Total gas estimate for redeeming all InitialReporter reporting fees.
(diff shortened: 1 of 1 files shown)


src/reporting/claim-reporting-fees.js
```diff
@@ -36,7 +36,7 @@ var PARALLEL_LIMIT = require("../constants").PARALLEL_LIMIT;
  * @property {Array.<string>|null} redeemedDisputeCrowdsourcers Addresses of all successfully redeemed Dispute Crowdsourcers, as hexadecimal strings.  Not set if `p.estimateGas` is true.
  * @property {Array.<string>|null} redeemedInitialReporters Addresses of all successfully redeemed Initial Reporters, as hexadecimal strings.  Not set if `p.estimateGas` is true.
  * @property {gasEstimateInfo|null} gasEstimates Object containing a breakdown of gas estimates for all reporting fee redemption transactions. Not set if `p.estimateGas` is false.
- * @property {Array.<RPCError|Error>} errors Array of errors returned when attempting to make transactions or get gas estimates, indexed by contract address.
+ * @property {Array.<RPCError|Error>} failedTransactions Array of errors returned when attempting to make transactions or get gas estimates, indexed by contract address.
```
epheph: Should be an array of `{address, error}`

adrake33: Good catch! Will update.

PR #313, opened 2018-03-30
created script to get correct docker images based on augur-core version

bthaile:


bthaile:
docker pull is no longer needed. moving to using specific version based on tags.

PR #333, opened 2018-04-12
Add `getReportingFees` to augur.js

adrake33:
Now that `getReportingFees` has been added to augur-node, we should also create a function to call it from augur.js.


epheph:
Super 👍

PR #335, opened 2018-04-13
Modify input/output format for claimReportingFees function

adrake33:
Updating input params to more closely match what's returned by the augur-node function `getReportingFees`. Also changed from a single callback function to using `onSent`, `onFailed`, and `onSuccess`.


adrake33:
@bthaile We should wait to merge this at the same time as https://github.com/AugurProject/augur/pull/1403. Otherwise it is going to break things in the UI and unit tests.

PR #338, opened 2018-04-21
Tomhaile/ch9280/add code to deployment scripts for creating

bthaile:
added script to help force market to finalization
added script to force a number of dispute rounds
added script to set timestamp (better than push-time)
added script to push timestamp

See >flash -h for details

### DIFF
--- a/scripts/flash/force-dispute.js
+++ b/scripts/flash/force-dispute.js
@@ -1,0 +1,111 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var chalk = require("chalk");
+var getTime = require("./get-timestamp");
+var setTimestamp = require("./set-timestamp");
+var repFaucet = require("../rep-faucet");
+var getPayoutNumerators = require("./get-payout-numerators");
+var doInitialReport = require("./do-initial-report");
+var doMarketContribute = require("./do-market-contribute");
+var displayTime = require("./display-time");
+
+function help() {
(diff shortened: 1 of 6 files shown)


scripts/flash/force-dispute.js
```diff
@@ -0,0 +1,111 @@
...
+function forceDispute(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
+  var marketId = args.opt.marketId;
+  var rounds = parseInt(args.opt.rounds, 10);
+  repFaucet(augur, 10000000, auth, function (err) {
+    if (err) return callback(err);
+    augur.markets.getMarketsInfo({ marketIds: [marketId] }, function (err, marketsInfo) {
```
adrake33: Should be checking `err` here (and a few lines below).

scripts/flash/force-finalize.js
```diff
@@ -0,0 +1,62 @@
...
+
+function forceFinalize(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
+  var marketId = args.opt.marketId;
+  repFaucet(augur, 10000000, auth, function (err) {
+    if (err) return callback(err);
+    augur.markets.getMarketsInfo({ marketIds: [marketId] }, function (err, marketsInfo) {
```
adrake33: Also should be checking `err` here (and a few lines below).

PR #344, opened 2018-04-27
Add unit tests for `augur.reporting.claimreportingfees`

adrake33:
NOTE: There is a fair amount of duplicated code in these tests. I started out writing the tests for the scenario where a finalized forked parent universe exists, and then copied/modified these tests for the scenarios where a non-finalized forked parent universe exists and no parent universe exists.  This was done in the interest of time because I did not know how to create shared tests.  I think I know how to do this now, but since these are working, I figure they can be merged, and I can work on creating shared tests as a separate task.


test/unit/reporting/claim-reporting-fees-no-fork.js
```diff
@@ -0,0 +1,552 @@
...
+"use strict";
+
+var assert = require("chai").assert;
+var assign = require("lodash.assign");
+var BigNumber = require("bignumber.js");
+var proxyquire = require("proxyquire").noPreserveCache();
+var noop = require("../../../src/utils/noop");
+var sinon = require("sinon");
+
+describe.only("reporting/claim-reporting-fees", function () {
```
bthaile: Probably want to remove .only

adrake33: Ah, yeah. Thanks, I forgot to remove that before committing.

PR #345, opened 2018-04-27
Tomhaile/ch9532/standardize flash scripts parameters

bthaile:
Use options parser throughout flash scripts. Beefed up help.

### DIFF
--- a/scripts/flash/designated-report.js
+++ b/scripts/flash/designated-report.js
@@ -25,7 +25,6 @@ function designateReport(augur, args, auth, callback) {
   }
   var amount = 10000;
   var marketId = args.opt.marketId;
-  var userAuth = auth;
   var outcome = args.opt.outcome;
   var invalid = args.opt.invalid;
   getRepTokens(augur, amount, auth, function (err) {
@@ -40,10 +39,6 @@ function designateReport(augur, args, auth, callback) {
         callback("Error");
       }
       var marketPayload = { tx: { to: marketId } };
-      // make sure we have the correct account to report with
-      if (userAuth.address !== market.designatedReporter && auth.address === market.designatedReporter) {
-        userAuth = auth;
-      }
       augur.api.Market.getEndTime(marketPayload, function (err, endTime) {
         console.log(chalk.red.dim("Market End Time"), chalk.red(endTime));
         getTime(augur, auth, function (err, timeResult) {
@@ -58,7 +53,7 @@ function designateReport(augur, args, auth, callback) {
               return callback(err);
             }
             var payoutNumerators = getPayoutNumerators(market, outcome, invalid);
-            doInitialReport(augur, marketId, payoutNumerators, invalid, userAuth, function (err) {
+            doInitialReport(augur, marketId, payoutNumerators, invalid, auth, function (err) {
               if (err) {
                 return callback("Initial Report Failed");
               }
--- a/scripts/flash/designated-report.js
+++ b/scripts/flash/designated-report.js
@@ -9,13 +9,25 @@ var setTimestamp = require("./set-timestamp");
 var getRepTokens = require("./get-rep-tokens");
 var doInitialReport = require("./do-initial-report");
 var getPayoutNumerators = require("./get-payout-numerators");
-var getPrivateKeyFromString = require("../dp/lib/get-private-key").getPrivateKeyFromString;
+
+function help() {
+  console.log(chalk.red("This command will move time to the designate report time fram and do an initial-report"));
+  console.log(chalk.red("FYI: Scalar markets outcome is the value between min and max, use market-info for details"));
+}
 
 /**
  * Move time to Market end time and do initial report
  */
-function designateReportInternal(augur, marketId, outcome, userAuth, invalid, auth, callback) {
+function designateReport(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
   var amount = 10000;
+  var marketId = args.opt.marketId;
+  var userAuth = auth;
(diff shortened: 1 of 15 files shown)


scripts/flash/dispute-contribute.js
```diff
@@ -91,61 +97,30 @@ function disputeContributeInternal(augur, marketId, outcome, amount, disputerAut
...
+      }
+      amount = balances.reputation;
       console.log(chalk.yellow.dim("amount"), amount);
-      disputeContributeInternal(augur, marketId, outcomeId, amount, userAuth, invalid, auth, callback);
-    }
+      disputeContributeInternal(augur, marketId, outcome, amount, userAuth, invalid, auth, callback);
+    });
+  } else {
+    console.log(chalk.yellow.dim("amount"), amount);
+    disputeContributeInternal(augur, marketId, outcome, amount, userAuth, invalid, auth, callback);
```
adrake33: Recommend moving these 2 lines outside the if/else block and removing the else block.

scripts/flash/fork.js
```diff
@@ -7,14 +7,24 @@ var getTime = require("./get-timestamp");
...
  * Move time to Market end time and do go to a fork
  */
-function forkInternal(augur, marketId, userAuth, auth, callback) {
+function fork(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
+  var marketId = args.opt.marketId;
+  var userAuth = auth;
```
adrake33: `userAuth` seems redundant. I think you could just use auth everywhere.

scripts/flash/designated-report.js
```diff
@@ -9,13 +9,25 @@ var setTimestamp = require("./set-timestamp");
...
  */
-function designateReportInternal(augur, marketId, outcome, userAuth, invalid, auth, callback) {
+function designateReport(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
   var amount = 10000;
+  var marketId = args.opt.marketId;
+  var userAuth = auth;
```
adrake33: It seems odd to set `userAuth` = `auth` here when later in the code there's this line:

```
if (userAuth.address !== market.designatedReporter && auth.address === market.designatedReporter) {
        userAuth = auth;
}
```

scripts/flash/index.js
```diff
@@ -30,55 +31,101 @@ var NETWORKS = ["aura", "clique", "environment", "rinkeby", "ropsten"];
...
+    opts: {
+      help: {flag: true, short: "h", help: "This help, dispute this market all the way to fork" },
+      marketId: { required: true, short: "m", help: "Required market id" },
+    },
   },
   "approval": {
     method: approval,
-    opts: null,
+    opts: {
+      help: {flag: true, short: "h", help: "This help" },
```
adrake33: Why does `help` always start with "This help"?  I think this one and the next one should have further descriptions.

scripts/flash/index.js
```diff
@@ -232,38 +255,31 @@ if (require.main === module) {
...
-    help();
-    process.exit();
+
+  if (args.opt.help) {
+    if (!method) {
+      help();
+      process.exit();
+    } else {
+      options.help(method.opts);
+      process.exit();
```
adrake33: This line can be moved out of here below the if/else block.

scripts/flash/initial-report.js
```diff
@@ -7,13 +7,25 @@ var getTime = require("./get-timestamp");
...
  */
-function initialReportInternal(augur, marketId, outcome, userAuth, invalid, auth, callback) {
+function initialReport(augur, args, auth, callback) {
+  if (args === "help" || args.opt.help) {
+    help();
+    return callback(null);
+  }
+  var marketId = args.opt.marketId;
+  var outcome = args.opt.outcome;
+  var userAuth = auth;
```
adrake33: Recommend just using `auth` here as well.

scripts/flash/push-time.js
```diff
@@ -3,80 +3,16 @@
...
-            modTimeBy = 1; // seconds
-          }
-          var totalMovement = modTimeBy * digit;
-          var newTimestamp = subtraction ? timestamp - totalMovement : timestamp + totalMovement;
-          setTimestamp(augur, newTimestamp, timeResult.timeAddress, auth, callback);
-        }
-      });
-    });
-  }
+function pushTime(augur, args, auth, callback) {
```
adrake33: Are you passing in `augur`, `args`, & `auth` here for consistency? They're not used in this function.

scripts/flash/show-initial-reporter.js
```diff
@@ -5,10 +5,20 @@
...
+function help() {
+  console.log(chalk.red("Show initial reporter account for market"));
+  console.log(chalk.red("Displays all 0s if market hasn't been reported on"));
+}
+
 /**
  * Move time to Market end time and do initial report
  */
-function showInitialReporterInternal(augur, marketId, auth, callback) {
+function showInitialReporter(augur, args, auth, callback) {
```
adrake33: `auth` param doesn't get used here.  I didn't check every function in this PR for unused params, but if this is something that should be corrected, I recommend double-checking the other functions in this PR as well.

scripts/flash/push-time.js
bthaile: yes, that is the pattern. every command gets passed these options.

scripts/flash/show-initial-reporter.js
bthaile: Every command gets passed these parameters/options. It's like an interface if we were using typescript.

PR #346, opened 2018-04-28
Can't see actual error in augur js connection

tinybike:
https://app.clubhouse.io/augur/story/9540/can-t-see-actual-error-in-augur-js-connection-error-output


tinybike:
This is a pretty trivial fix, so I'm going to go ahead and merge.

PR #364, opened 2018-05-08
Properly clone the object so it does not get modified when configurat…

epheph:
…ion object does

Also, check for blank strings and don't bother adding to array

### DIFF
--- a/test/unit/connect/create-configuration.js
+++ b/test/unit/connect/create-configuration.js
@@ -8,6 +8,9 @@ var createConfiguration = require("../../../src/connect/create-configuration");
 describe("connect/create-configuration", function () {
   var test = function (t) {
     it(t.description, function () {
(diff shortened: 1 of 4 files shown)


test/unit/connect/create-configuration.js
```diff
@@ -167,4 +171,29 @@ describe("connect/create-configuration", function () {
...
+    assertions: function (configuration) {
+      assert.deepEqual(configuration, {
+        http: "http://somewhere:1234",
+        ws: "",
+        ipc: null,
+        api: { events: null, functions: null },
+        contracts: { 3: { contract1: "0xc1", contract2: "0xc2" } },
+        httpAddresses: ["http://somewhere:1234"],
+        wsAddresses: [],
+        ipcAddresses: [],
```
epheph: The gist of this test is that wsAddresses here is not `[""]`

PR #368, opened 2018-05-10
Separate claiming reporting fees from nonforked markets and forked market into separate functions

adrake33:
NOTE: This PR also modifies the logic for claiming initial reporters on non-forked markets.  Previously, `InitialReporter.redeem` was getting called on all initial reporters passed in to `claimReportingFees`.  However, Scott pointed out that we should not be calling `InitialReporter.redeem` on non-forked markets where `Market.disavowCrowdsourcers`had been called, because when this function is called, the initial reporter is disavowed, but then reattached to the market.  Therefore, the initial reporter should only be redeemed once such markets have been migrated.  Now, it is only initial reporters belonging to non-forked, finalized markets that have `InitialReporter.redeem` called on them (in the case where a forked market exists in the current universe). If no forked market exists in the current universe, the all initial reporters for non-forked markets will have `InitialReporter.redeem` called on them.


src/reporting/claim-reporting-fees-forked-market.js
```diff
@@ -0,0 +1,237 @@
...
+              nextContract();
+            },
+            onFailed: function () {
+              failedTransactions.crowdsourcerForkAndRedeem.push(contract.address);
+              // console.log("Failed to forkAndRedeem crowdsourcer", contract.address);
+              nextContract();
+            },
+          }));
+        } else {
+          api().DisputeCrowdsourcer.redeem(assign({}, payload, {
```
epheph: This is for initialReporter yea? I believe it should also fork?

epheph: nevermind about comment, i wasn't lining up my if blocks right :)

src/reporting/claim-reporting-fees-forked-market.js
```diff
@@ -0,0 +1,237 @@
...
+            onSent: function () {},
+            onSuccess: function (result) {
+              if (p.estimateGas) {
+                result = new BigNumber(result, 16);
+                gasEstimates.crowdsourcerRedeem.push({address: contract.address, estimate: result});
+                gasEstimates.totals.crowdsourcerRedeem = gasEstimates.totals.crowdsourcerRedeem.plus(result);
+              } else {
+                successfulTransactions.crowdsourcerRedeem.push(contract.address);
+              }
+              // console.log("Redeemed crowdsourcer", contract.address);
```
epheph: Remove debugging comments?

src/reporting/claim-reporting-fees-nonforked-markets.js
```diff
@@ -0,0 +1,279 @@
...
+    totals: {
+      disavowCrowdsourcers: new BigNumber(0),
+      feeWindowRedeem: new BigNumber(0),
+      crowdsourcerRedeem: new BigNumber(0),
+      initialReporterRedeem: new BigNumber(0),
+      all: new BigNumber(0),
+    },
+  };
+
+  if (p.forkedMarket) {
```
epheph: This comes down to the question again of how we communicate the need to disavow, will we always return a forkedMarket in the object, or put information in with the market?

src/reporting/claim-reporting-fees-forked-market.js
adrake33: Yeah, I considered doing that, but I think I'll leave them in for now in case we want to use them later for debugging.

src/reporting/claim-reporting-fees-nonforked-markets.js
adrake33: In the spec doc (https://docs.google.com/document/d/1PBa8F2_ZlD3YTve7Dy9PUGM8h_A_qeJKwRfn1PK6Gy4/edit), I wrote:

```
// The forked market object should always be returned when a forked market exists in the current universe, even if the user does not have any claimable fees in it. This is because `claimReportingFeesNonforkedMarkets` needs to know if a non-finalized forked market exists so it can call Market.disavowCrowdsourcers on non-finalized non-forked markets if the forked market is not finalized.
```
However, in looking through the `getReportingFees` function, I don't think we're actually including it if there are no crowdsourcers/initial reporters that the specified user has unclaimed fees in. Is this something you'd like me to add, or would it be easier for you to do it since you're already modifying the `getReportingFees` code?

adrake33: Right, I added the `needsFork` property to the section of the doc containing the format of the returned data. However, that only applies to reporting participants in the forked market, correct? This function only operates on reporting participants in non-forked markets, so I don't think I can use that.

epheph: hmm, i suppose we could add it as an attribute to markets?

epheph: If you want to do it that way, always return forked market details, that's fine with me too. I just thought it might be simpler for your implementation if augur-node tracked the requirements of various markets/participants and what they need

adrake33: I think it makes sense to always return forked market details. The Portfolio: Reporting page will also need to know if a forked market exists in order to determine whether to show the forked market card under the Claim button in the UI.

PR #369, opened 2018-05-10
Tinybike/ch10130/sending eth you shouldn t have to send in

tinybike:
Should resolve [ch10130](https://app.clubhouse.io/augur/story/10130/sending-eth-you-shouldn-t-have-to-send-in-tx).  Requires https://github.com/AugurProject/augur-node/pull/331


JohnDanz:
Lookin' good, and it helped me surface an issue in the UI where we were sending erroneous share balances sometimes. merging...

PR #371, opened 2018-05-14
add noPush to dispute contribute so we can dispute without moving time

bthaile:
added --noPush to dispute contribute so we can dispute without moving time. 

makes it easier to combine push-timestamp and dispute-contribute.


scripts/flash/dispute-contribute.js
```diff
@@ -66,8 +91,6 @@ function disputeContributeInternal(augur, marketId, outcome, amount, disputerAut
                 console.log(chalk.red(err));
                 return callback(err);
               }
-              var payoutNumerators = getPayoutNumerators(market, outcome, invalid);
-              var attoREP = speedomatic.fix(amount, "hex");
```
adrake33: Why were these lines removed? Were these variables unused?

PR #372, opened 2018-05-15
Add transfer assets flash

bthaile:
updated get-balance to show fee window address.

added new flash script to transfer ETH or REP assets.

### DIFF
--- a/scripts/flash/transfer-assets.js
+++ b/scripts/flash/transfer-assets.js
@@ -1,0 +1,71 @@
+#!/usr/bin/env node
+
+"use strict";
+
+var chalk = require("chalk");
+var theGetBalance = require("../dp/lib/get-balances");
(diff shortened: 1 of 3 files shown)


scripts/flash/transfer-assets.js
```diff
@@ -0,0 +1,68 @@
...
+      },
+      onFailed: function (err) {
+        console.log(chalk.red(err));
+        return callback(err);
+      },
+    });
+  } else {
+    var message = "Asset type unknown";
+    console.log(chalk.red(message));
+    return callback(message);
```
adrake33: Recommend passing `message` back as `new Error(message)` instead of as a string.

adrake33: Never mind, looks like we do `return callback(new Error(message))` in the code, but `return callback(message)` is more common.

PR #373, opened 2018-05-15
Subscribe to blockstream event logs by contract, not by event

epheph:


src/events/add-filter.js
```diff
@@ -1,17 +1,33 @@
...
+  }
+  var contractAddress = contracts[contractName];
+  addSubscription(contractAddress, blockStream.addLogFilter({
     address: contractAddress,
-    topics: [eventAbi.signature],
+    topics: [Object.keys(eventBySignature)],
   }), function (message) {
-    parseLogMessage(contractName, eventName, message, eventAbi.inputs, onMessage);
+    var event = eventBySignature[message.topics[0]];
+    if (!eventName) return console.warn("Ethereum Node provided data we did not request", contractAddress, message.topics[0]);
```
MicahZoltu: It feels like the `parseLogMessage` line should be resilient to receiving calls that it doesn't understand.  If it is, then there is no need for this line.  If it isn't, then I worry a little about accidentally letting through a wrong log.

src/events/start-blockchain-event-listeners.js
```diff
@@ -28,15 +28,10 @@ function startBlockchainEventListeners(eventCallbacks, startingBlockNumber, onSe
   blockStream.subscribeToOnLogAdded(subscriptions.onLogAdded);
   blockStream.subscribeToOnLogRemoved(subscriptions.onLogRemoved);
   async.forEachOf(eventCallbacks, function (callbacks, contractName, nextContract) {
```
MicahZoltu: Is this loop still necessary?  If we are only ever adding one filter, it feels like we shouldn't be looping here?

epheph: There's still 3. In the version you discussed in CH, it won't be necessary

src/events/add-filter.js
epheph: If we're going to go to the single getLogs version, it seems like it might be more important to have a "hey, this matched but shouldn't have" warning?

src/events/subscriptions.js
```diff
@@ -7,34 +7,31 @@ var initialState = { block: {} };
...
 
 module.exports.removeSubscription = function (token) {
   subscriptions = Object.keys(subscriptions).reduce(function (p, contractAddress) {
-    Object.keys(subscriptions[contractAddress]).forEach(function (eventSignature) {
-      if (subscriptions[contractAddress][eventSignature].token !== token) {
-        p[contractAddress][eventSignature] = subscriptions[contractAddress][eventSignature];
-      }
-    });
+    if (subscriptions[contractAddress].token !== token) {
+      p[contractAddress] = subscriptions[contractAddress];
```
MicahZoltu: Unrelated to this PR, `p` is a _terrible_ variable name.

src/events/add-filter.js
MicahZoltu: Yeah, I guess my question is really, "should that go here or down in `parseLogMessage` or deeper?  I tend to prefer having things be as resillient as possible, which usually ends up resulting in errors being handled pretty deep rather than high up.

I'm not familiar with this code, so certainly use your discretion as I may not be understanding the overarching patterns or intended usages.

src/events/subscriptions.js
epheph: Well, when we go to 1, we don't need it anymore!

src/events/add-filter.js
epheph: We can make parseLogMessage more resilient, too, but it does seem a bit awkward to enter it with undefined eventName, abi, and callback and let it notice. Especially since we are recognizing the risk of that happening by combining the 3 contracts into a single getLogs. It seems like we would be bothering the parser with the implementation of how we retrieved the logs

src/events/subscriptions.js
tinybike: Ah, yes I have a bad habit of using `p` (...short for `previous`) for reduce's accumulator parameter... please feel free to rename these to something more descriptive when you encounter them :grimacing:

src/events/start-blockchain-event-listeners.js
tinybike: Not sure what other changes are queued up, but FWIW, `addFilter` currently calls `blockStream.addLogFilter` and sets up the callback (in augur.js) for each event.

epheph: The change here is to to not do that by event, but by contract, to significantly reduce the number of getLogs requests per block. Micah is suggesting further changing to do it just once per block

epheph:
@tinybike mentioned that this might be even better handled inside blockstream, deferring to it to group them together in the most optimal way. But then i wonder if lumping all the cross-contract getLogs together and parsing out is a good idea since it might not be a good solution for other projects where there could be more signature collisions? (Approval/Transfer/other standard methods).

(Then i wonder if we should just be using the block's bloom to be smarter about it, probably the most optimal of all solutions.)

I think we should do either what @MicahZoltu said, and group them all into a single listen inside augur.js, or what @tinybike said and delegate the optimization to inside of blockstream, but not both?

Anyway, this is better than it was, and while we're deciding, we can benefit from this new version (I'm shipping augur.js right now). It doesn't change any interfaces between them, so it's still just as easy to make correct decision down the road.

PR #374, opened 2018-05-16
added fee window to get balances of the fee window of the market

bthaile:
If market has fee window get the cash and eth balance.

### DIFF
--- a/scripts/flash/get-market-balance.js
+++ b/scripts/flash/get-market-balance.js
@@ -21,6 +21,43 @@ function showWinningBalance(augur, marketId, address, callback) {
   });
 }
 
+function showCashBalance(augur, address, label, callback) {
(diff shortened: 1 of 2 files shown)


scripts/flash/get-market-balance.js
```diff
@@ -17,10 +17,53 @@ function help() {
...
+    console.log(chalk.green.dim("Total ETH balance:"), chalk.green(totalCashBalance));
+    callback(null);
+  });
+}
+
+function showBalances(augur, address, label, callback) {
+  console.log(chalk.yellow.dim(label), chalk.yellow(address));
+  showCashBalance(augur, address, label, function (err) {
+    if (err) {
+      console.log(chalk.red(err));
```
adrake33: Shouldn't this also do `return callback(err);` here?

adrake33: Never mind, I think this is ok.

PR #376, opened 2018-05-17
Fix time moving backwards

bthaile:
Guard against moving time backwards when forcing finalization or just regular finalization.


scripts/flash/force-finalize.js
```diff
@@ -41,22 +41,33 @@ function forceFinalize(augur, args, auth, callback) {
...
-            if (err) {
-              console.log(chalk.red(err));
-              return callback(err);
-            }
-            var priceOrOutcome = market.marketType === "scalar" ? market.minPrice : 0;
-            var payoutNumerators = getPayoutNumerators(market, priceOrOutcome, false);
+          displayTime("Current Time", timeResult.timestamp);
+          var priceOrOutcome = market.marketType === "scalar" ? market.minPrice : 0;
+          var payoutNumerators = getPayoutNumerators(market, priceOrOutcome, false);
+          if (timeResult.timestamp > endTime) {
```
adrake33: Should this be `parseInt(timeResult.timestamp, 10)` like above?

PR #394, opened 2018-05-19
noticed that in higher dispute rounds all of the users rep wasn't use…

bthaile:
Needed when not sending in amount just use accounts full balance.

### DIFF
--- a/scripts/flash/dispute-contribute.js
+++ b/scripts/flash/dispute-contribute.js
@@ -87,7 +87,7 @@ function disputeContribute(augur, args, auth, callback) {
             augur.api.FeeWindow.getStartTime(feeWindowPayload, function (err, feeWindowStartTime) {
               if (err) {
                 console.log(chalk.red(err));
-                callback("Could not get Fee Window");
+                return callback("Could not get Fee Window");
               }
 
               getTime(augur, auth, function (err, timeResult) {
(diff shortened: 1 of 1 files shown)


scripts/flash/dispute-contribute.js
```diff
@@ -45,76 +44,96 @@ function disputeContribute(augur, args, auth, callback) {
...
+              console.log(chalk.red("feeWindowId has not been created"));
+              return callback("Market doesn't have fee window, need to report");
+            }
+            console.log(chalk.yellow("Market Fee Window"), chalk.yellow(feeWindowId));
+            var feeWindowPayload = { tx: { to: feeWindowId } };
+            augur.api.FeeWindow.getStartTime(feeWindowPayload, function (err, feeWindowStartTime) {
               if (err) {
                 console.log(chalk.red(err));
-                return callback(err);
+                callback("Could not get Fee Window");
```
adrake33: I think this should be `return callback`

scripts/flash/dispute-contribute.js
```diff
@@ -45,76 +44,96 @@ function disputeContribute(augur, args, auth, callback) {
...
+                        if (err) {
+                          return callback("Market contribute Failed");
+                        }
+
+                        console.log(chalk.green("Market contribute Done"));
+                        callback(null);
+                      });
+                    } else {
+                      console.log(chalk.red("Fee Window isn't active"));
+                      callback("Fee Window isn't active");
```
adrake33: This should also probably be `return callback`.

scripts/flash/dispute-contribute.js
bthaile: good call

scripts/flash/dispute-contribute.js
bthaile: just for convention, I agree. will fix

PR #395, opened 2018-05-19
Achapman/ch11120/for disavowinig crowdsourcers clicking claim

nuevoalex:


src/reporting/claim-reporting-fees-nonforked-markets.js
```diff
@@ -42,10 +44,17 @@ function redeemContractFees(p, payload, successfulTransactions, failedTransactio
...
   for (i = 0; i < p.nonforkedMarkets.length; i++) {
     for (var j = 0; j < p.nonforkedMarkets[i].crowdsourcers.length; j++) {
-      redeemableContracts.push({
-        address: p.nonforkedMarkets[i].crowdsourcers[j],
-        type: contractTypes.DISPUTE_CROWDSOURCER,
-      });
+      var crowdsourcerAddress = p.nonforkedMarkets[i].crowdsourcers[j];
+      var marketWasDisavowed = successfulTransactions.disavowCrowdsourcers.indexOf(p.nonforkedMarkets[i].marketId) !== -1;
+      if (p.estimateGas && marketWasDisavowed) {
+        gasEstimates.crowdsourcerRedeem.push({address: crowdsourcerAddress, estimate: CROWDSOURCER_REDEEM_ESTIMATE});
```
adrake33: I'm not sure I understand why this is being done here.  Wouldn't it make more sense to do this later in the code, such as the `onSuccess` function when `api().DisputeCrowdsourcer.redeem` is called?

nuevoalex: That function fails in this context since yiu cant redeem before disavowing. This code was failing for that reason.

nuevoalex:
Tom this PR isnt for the UI.

src/reporting/claim-reporting-fees-nonforked-markets.js
adrake33: Ah, ok I see.

PR #405, opened 2018-05-23
by default do not snap num ticks. purely a UI input nice-ity

nuevoalex:
The only reason to do this snap is to covert human input into a nice even amount. Everywhere else we need precision.

This fixes the current issue with buying out a cat 7 order book then selling held shares

Update: On reflection I believe the snapping behavior is entirely an error and am just removing it here (confirmed the bug is still fixed in the new version). Whatever purchase from converting from display amount to on chain _should_ then appear everywhere in the UI as the originally entered display amount, so this would only fix display issues if we were incorrectly converting to display somewhere.

### DIFF
--- a/src/utils/convert-display-amount-to-on-chain-amount.js
+++ b/src/utils/convert-display-amount-to-on-chain-amount.js
@@ -1,12 +1,8 @@
 "use strict";
 
-var BigNumber = require("bignumber.js");
 var speedomatic = require("speedomatic");
 
-function convertDisplayAmountToOnChainAmount(displayAmount, displayRange, numTicks, snapNumTicks) {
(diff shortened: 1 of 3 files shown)


src/utils/convert-display-amount-to-on-chain-amount.js
```diff
@@ -1,12 +1,8 @@
 "use strict";
 
-var BigNumber = require("bignumber.js");
 var speedomatic = require("speedomatic");
 
 function convertDisplayAmountToOnChainAmount(displayAmount, displayRange, numTicks) {
-  if (numTicks.eq(10002) || numTicks.eq(10003)) {
```
tinybike: Good find, I missed this one :(

PR #408, opened 2018-05-24
adjust sim trade to not filter by user

nuevoalex:
To accommodate the upcoming contract restriction removal on taking ones own orders


tinybike:
It looks like the relevant change has been merged into augur-core, so merging this as well.

PR #412, opened 2018-05-27
added labels for DR and market id

bthaile:
There has been some confusion to what address is what, added labels.


bthaile:
Need to add newline between markets. Don’t merge just yet.

PR #423, opened 2018-06-14
needed to have returns to finish execution

bthaile:
not super critical but needed to add returns to quickly return instead of running through.


scripts/flash/set-timestamp.js
```diff
@@ -11,7 +11,7 @@ function setTimestamp(augur, newTimestamp, address, auth, callback) {
   getTime(augur, auth, function (err, timeResult) {
     if (parseInt(timeResult.timestamp, 10) === parseInt(newTimestamp, 10)) {
       console.log(chalk.yellow.dim("Time already set"));
-      callback(null);
+      return callback(null);
```
epheph: This seems to be the only one that would make a difference?

PR #429, opened 2018-06-26
Augur core@1.0.0 14

epheph:


epheph:
Tests failing because we are depending on a typescript module that is coming from git because it hasn't been published to npm so it doesn't have access to tsc .

PR #439, opened 2018-07-06
Revert "Justinbarry/ch13255/only use web3transport if user specifically"

bthaile:
Reverts AugurProject/augur.js#437


bthaile:
I think we are good with Justin's PR.

PR #461, opened 2018-07-13
always try to do at least one trade

nuevoalex:
The "withLimit" trade functions have irregular behavior when they are prevented from actually filling orders.


bthaile:
Is this the sending 0 to contracts which cause issue fix?

PR #469, opened 2018-07-18
allow passing in numTick array in ###,### format

bthaile:
This change is to make setting nonsensical report easier.

https://app.clubhouse.io/augur/story/14054/ui-display-for-binary-cat-markets-that-receive-nonsensical-reports


adrake33:
Lgtm

PR #491, opened 2018-07-29
Uploaded contracts to Thunder testnet

tinybike:


tinybike:
No

tinybike:
As mentioned in the CH issue, the endpoint I'm connecting to is the hosted Thunder testnet node, i.e. it's not local and therefore can't be using IPC

tinybike:
@bthaile I just thought it made sense to use whatever was specified in the network object, rather than only selecting http for some reason.

PR #492, opened 2018-07-30
Added better error handling for startAugurNodeEventListeners

adrake33:


epheph:
@adrake33 this looks good and is an important check, but it only catches half of the invalid, if they pass in a bad function. Here is some augur-node work to begin fixing it on the augur-node side:

https://github.com/AugurProject/augur-node/pull/581

adrake33:
@epheph Good idea! I'll look over your PR and modify `startAugurNodeEventListeners` to take advantage of these changes.

PR #493, opened 2018-07-30
Create constants for augur node subscription events

adrake33:
As per @pgebheim's recommendation, I added a TS enum in index.d.ts and added the transpiled version to constants.js. I also alphabetized the constants to make them easier to find in the file. (Helps when comparing the existing constants with what we have in the docs.)


epheph:
@adrake33 resolve conflict and let's merge, then i'll fix up augur-node to use this exported enum

adrake33:
@epheph Sounds good. Conflict is now resolved.

PR #495, opened 2018-08-01
change network id's to be below 109

petong:
I *think* this should do it


epheph:
This is a step in the right direction (maybe? It's not a step in the wrong direction, for sure), as it will change the chainId, because we set the networkId and chainId to the same value in our geth nodes. It will not fully fix our issue, though, as chainId (for some insane reason) is NOT queryable from the eth node (😭 🤷‍♂️ ).
So somewhere in our transaction-creation libraries, there must be some static mapping of networkId to chainId in the ethereum client ? And we'll need to add 101=101, 102=102, etc there.  

An EIP to add chainId is merged but not implemented: https://github.com/ethereum/EIPs/issues/694

PR #501, opened 2018-08-03
Fix naming convention for SUBSCRIPTION_EVENT_NAMES & ORDER_STATE

adrake33:


index.d.ts
```diff
@@ -156,11 +156,11 @@ interface BlockSubscriptionCallbacks {
   onRemoved?: (block: Block) => void;
 }
 
-export enum OrderState {
-  All = "ALL",
-  Cancelled = "CANCELLED",
```
epheph: This will be a larger if this drops the "L". if we want to do it, we should line it up with another DB forced resync

adrake33: Oh, but isn't AN using "OrderState.Canceled", not "OrderState.Cancelled"?

pgebheim: We can, and should, do a migration to fix the DB instead of doing force re-syncs.

adrake33:
Discussed this further with @epheph and @pg, and we decided no further changes need to be made in order to use OrderState.CANCELED in augur-node, because this is already the format being used there.

PR #510, opened 2018-08-24
More event name constants

epheph:


index.d.ts
```diff
@@ -20,12 +19,14 @@ export enum SUBSCRIPTION_EVENT_NAMES {
...
   FEE_WINDOW_CLOSED = "FeeWindowClosed",
   FEE_WINDOW_CREATED = "FeeWindowCreated",
+  FEE_WINDOW_OPENED = "FeeWindowOpened",
   FEE_WINDOW_REDEEMED = "FeeWindowRedeemed",
   INITIAL_REPORT_SUBMITTED = "InitialReportSubmitted",
   INITIAL_REPORT_REDEEMED = "InitialReporterRedeemed",
   INITIAL_REPORT_TRANSFERRED = "InitialReporterTransferred",
   MARKET_CREATED = "MarketCreated",
   MARKET_STATE = "MarketState",
+  MARKET_MIGRATED = "MarketMigrated",
```
adrake33: Doesn't this also need a MARKET_FINALIZED like below?

epheph: fixed, thx

PR #517, opened 2018-09-12
Change getAllAugurLogs behavior to include callbacks, even for empty …

epheph:
…getLogs.

This will allow us to check if those blocks are 'null' and infer we have a parity warp problem

This change will be a two parter. First is to get augur.js, which calls back batches of logs, to start calling back the range of blocks it has processed, even if there were no logs in it. This is because parity will sometimes lie about logs not being present when it is in "warp" mode, so we if we just query for the blocks (which will return `null` if it is in this state) we will be able to infer we are impacted by this issue.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -232,6 +232,9 @@ export interface RpcInterface {
   setDebugOptions(debugOptions: {[debugOptionName: string]: boolean}): void;
 }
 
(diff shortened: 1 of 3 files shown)


epheph:
@justinbarry this is the PR for the augur-node PR, BlockRange

PR #521, opened 2018-09-21
Juliyasmith/ch15938/GitHub silent errors when using bad ethereum

unparalleled-js:
https://app.clubhouse.io/augur/story/15938/github-silent-errors-when-using-bad-ethereum-node


justinbarry:
Good find!

PR #529, opened 2018-10-01
Implement getGasPrice override as module-level global

epheph:
Eventually refactor augur.js to be a real class

```
augur.getGasPrice = (callback) => { callback(7500000000)}
```


bthaile:
my function never gets called in order to pass in gasPrice. 
see augur-ui `phoebemirman/ch14649/github-give-edge-users-ability-to-change` branch
in `src/modules/app/actions/register-user-defined-gasPrice-function.js`, augur.getGasPrice function is set.

PR #547, opened 2018-11-07
Change to how event callbacks are dispatched and how logs are filtered

epheph:
Log filtering is now done with a single filter object, instead of one per contract
Based on this, refactored how we look up the results and the map required to perform this lookup
Changes the arguments and required data structures for startBlockchainEventListeners, should be major version bump


ryanberckmans:
Travis:
image

src/events/start-blockchain-event-listeners.js
```diff
@@ -1,41 +1,83 @@
...
+          eventName: eventName,
+          abi: eventAbi.inputs,
+        };
+        topics.add(eventAbi.signature);
+      }
+    }
+  }
+  const filter = {
+    address: addresses,
+    topics: [Array.from(topics)],
```
ryanberckmans: Double checked what I'm sure you already knew: augur.js supports `Array.from` via build using babel/polyfill.

src/events/start-blockchain-event-listeners.js
```diff
@@ -1,41 +1,83 @@
...
+  return compact(parsedLogs);
+}
+
+// Iterates through eventCallbacks, generates eth filter and event signature mapping
+function getEventsDetails(eventsToSubscribe, activeContracts) {
+  var eventByAddressAndSignature = {};
+  const addresses = [];
+  const topics = new Set();
+  for (const contractName in eventsToSubscribe) {
+    if (Object.prototype.hasOwnProperty.call(eventsToSubscribe, contractName)) {
```
ryanberckmans: Perhaps `if (!Object.prototype.hasOwnProperty.call(eventsToSubscribe, contractName)) { continue; }` to lower cyclomatic complexity of inner loop

epheph: We use eslint rule `no-continue`

PR #548, opened 2018-11-10
raise log parsing errors instead of continuing and just logging

nuevoalex:
In the event this exception occurs we would effectively just ignore the log. So one could imagine if we somehow hit this case and received for example no topics in our log response data it would just continue processing the other logs and never process the malformed log.

Letting this exception get raised will crash the process and prevent processors from potentially missing a log.

### DIFF
--- a/src/events/get-all-augur-logs.js
+++ b/src/events/get-all-augur-logs.js
@@ -51,8 +51,8 @@ function getAllAugurLogs(p, batchCallback, finalCallback) {
           try {
             return parseLogMessage(contractName, eventName, log, eventsAbi[contractName][eventName].inputs);
           } catch (exc) {
-            console.error("parseLogMessage error", exc);
             console.log(contractName, eventName, log, eventsAbi[contractName], chunkOfBlocks);
+            throw new Error(exc);
           }
         }
       });


src/events/get-all-augur-logs.js
```diff
@@ -51,8 +51,8 @@ function getAllAugurLogs(p, batchCallback, finalCallback) {
           try {
             return parseLogMessage(contractName, eventName, log, eventsAbi[contractName][eventName].inputs);
           } catch (exc) {
-            console.error("parseLogMessage error", exc);
             console.log(contractName, eventName, log, eventsAbi[contractName], chunkOfBlocks);
+            throw new Error(exc);
```
epheph: This is within an async callback; we have a `finalCallback`, perhaps we should pass an error back to it here? Until we get this turned into an async function, catching this error might be hard

pgebheim:
@nuevoalex Want us to do this still for V1 or should we just note it for V2?

justinbarry:
@pgebheim @nuevoalex This is an augur.js specific issue. Closing accordingly. Will create a new issue for discussion in V2.

PR #557, opened 2018-11-29
needed to subtract taken shares from shares avaialble when user is us…

bthaile:
…ing shares to take order

https://github.com/AugurProject/augur-ui/issues/2498

Needed to subtract taken shares from shares available in UI's call to simulate-trade.


bthaile:
Unit tests need to be fixed and I found an issue with my code change.

PR #571, opened 2019-03-07
Add option to connect to augur-node via http

justinbarry:
Requires an ethrpc release:
https://github.com/ethereumjs/ethrpc/commit/ea93c5d1f1426328d9399a30cc0f6f059a7459a8


justinbarry:
@bthaile No. This is just for AUGUR_NODE to deal with environments that do not have websockets available.

